PR #1, opened 2010-12-03
Makes concat work on Solaris.     

saysjonathan:
add a -g flag to concatfragments.sh to disable GNU extensions to find, sort and xargs; makes concat work on Solaris.

Add a $gnu parameter to make use of the aforementioned -g flag, and a $order parameter to make use of the -n flag in concatfragments.sh


hunner:
Rebased commit: 656c771ed50bcec616fac3a85a3af9b87ef89b3c

Oddly the `merge_commit_sha` is blank and the merge event is only a "referenced" event and there is no "merged" event. Less weird than #2 but still weird.

PR #5, opened 2011-04-20
updated to better support dash

TimidRobot:
Dash (default /bin/sh on Debian and Ubuntu) does not support 'echo -e'.  That invocation of echo is not POSIX compliant: https://bugs.launchpad.net/ubuntu/+source/dash/+bug/72167


izubalando:
#`

PR #230, opened 2014-09-13
set group of fragments to gid

duritong:
gid has now been part of facter for a long time, it is safe to use
it now.

### DIFF
--- a/manifests/fragment.pp
+++ b/manifests/fragment.pp
@@ -72,8 +72,9 @@ define concat::fragment(
   $safe_target_name = regsubst($target, '[/:\n]', '_', 'GM')
   $concatdir        = $concat::setup::concatdir
   $fragdir          = "${concatdir}/${safe_target_name}"
-  $fragowner            = $concat::setup::fragment_owner
-  $fragmode             = $concat::setup::fragment_mode
+  $fragowner        = $concat::setup::fragment_owner
+  $fraggroup        = $concat::setup::fragment_group
+  $fragmode         = $concat::setup::fragment_mode
 
   # The file type's semantics are problematic in that ensure => present will
   # not over write a pre-existing symlink.  We are attempting to provide
@@ -110,11 +111,10 @@ define concat::fragment(
     crit('No content, source or symlink specified')
   }
 
-  # punt on group ownership until some point in the distant future when $::gid
-  # can be relied on to be present
   file { "${fragdir}/fragments/${order}_${safe_name}":
     ensure  => $safe_ensure,
     owner   => $fragowner,
+    group   => $fraggroup,
     mode    => $fragmode,
     source  => $source,
     content => $content,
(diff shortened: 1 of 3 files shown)


sathieu:
any new on this PR?

jhoblitt:
I'm hesitant to assume `$::gid` is universally present but setting an internal constraint on `$::facterversion` seems like a lot of complexity for little net functionality.

@hunner Is there a [planned] way to place a dependency on the facter version via `metadata.json`?

Also, since this PR modifies the public API the README documentation would need to be updated.

duritong:
I had a quick look at the README and so far only API deprecations are documented. This change doesn't deprecate anything it adds a new feature. If you tell me where you like to have changes like this document, I can adapt the PR

duritong:
btw: I rebased it into current master

mhaskel:
Hi @duritong,

Thanks for the contribution! It looks like there aren't actually any docs updates needed since you didn't add any parameters that are exposed through a class/defined type.

This fact was added in facter 2.2, so I think we still need to account for it not being present. Could you wrap the new parameter in something like [this](https://github.com/puppetlabs/puppetlabs-concat/blob/master/manifests/init.pp#L202) and explicitly set `$fragment_group` to `undef` if the fact isn't defined?

Thanks!

duritong:
I updated my pull request as discussed

mhaskel:
great, thanks @duritong !

jhoblitt:
uh, this PR changes the public API...

duritong:
@jhoblitt I specifically asked you where to document it and later @mhaskel agreed with me that we didn't add a new parameter.

So if you want me to document that change, you must be more specific and verbose of what should be added. Thanks.

mhaskel:
@jhoblitt this only adds in-class variables, no new parameters.

jhoblitt:
@mhaskel  `concat::fragment::fraggroup` is not a parameter?

jhoblitt:
Ignore me -- my case of the flu has been talking.  This is :+1:

mhaskel:
feel better!

sodabrew:
Hi, this means I need Facter 2.0 or higher on all nodes. What's the purpose of this functionality in the first place? I never want the ownership of the concatfragments.sh to be anything other than root.

duritong:
It just brings everything in line. Owner and mode are already managed, so why not group?

Additionally, if you don't manage the group, group ownership is in kind of a limbo state, as it's not sure that the group will really be root. Especially, if you copy fragments from a source.
And as you might see from the deprecation messages, you can't actually set the owner, group nor mode it has no effect: https://github.com/puppetlabs/puppetlabs-concat/blob/0aa6c635c13e5978f38000e2f6d64b4796a02b4a/manifests/fragment.pp#L48-L56

And last but not least: We can't hardcode it to root, because this would make the concat module unusable if you run puppet as a unprivileged user. There are tons of valid use cases for that and there is no reason that the concat module requires root to be used.

sodabrew:
Thanks for the explanation. Looks like it's doing the right thing for me since #280 as a fallback for not having the gid fact on my older nodes.

PR #243, opened 2014-11-13
Support running a validation command on the destination file.

jmkeyes:
This patch allows a user to pass an optional `validate_cmd` parameter to a `concat` resource, which in turn will run a command to validate the output of the file before the module replaces it. It follows the same semantics as the `validate_cmd` parameter to the `file` resource.

Example:
```puppet
concat { '/etc/httpd/conf/httpd.conf':
  ensure       => file,
  validate_cmd => '/usr/sbin/apache2 -t -f %',
  owner        => 'apache2',
  group        => 'apache2',
  mode         => '0644'
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,13 @@ Ensure there's a newline at the end of the fragments.
 - ensure_newline => true
 - ensure_newline => false
 
+#####`validate_cmd`
+Ensure the destination file passes the following validation command.
+
+######Example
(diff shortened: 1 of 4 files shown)


mhaskel:
Thanks for the contribution @jmkeyes !

igalic:
@antaflos: ^ is this what you wanted?

PR #246, opened 2014-11-14
MODULES-1456 - make sure ruby is in path on PE

mhaskel:
Ruby is not in the path for all PE installations, so make sure the
concat script is run with ruby in the path.


joshbeard:
Tested and can vouch for functionality on AIX 6.1, AIX 7.1, and RHEL 6 :+1:

PR #257, opened 2015-01-15
Fix warning when generating files on windows machines

josecastroleon:
When generating a concatenated file on windows machines, the concat file shows a warning that the file should be generated with source_permissions => ignore. This patch tries to avoid that by specified a ignore default for windows machines.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -136,6 +136,11 @@ define concat(
     false => '',
   }
 
+  $source_permissions = $::operatingsystem ? {
+    'windows' => ignore,
+    default   => use,
+  }
+
   File {
     backup  => false,
   }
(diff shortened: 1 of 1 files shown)


cyberious:
Hi @josecastroleon, thanks for the PR.  What version of puppet are you using to get these warnings?  Is there a ticket for it?

josecastroleon:
We are using right now 3.4.3

cyberious:
I like the start of this, would prefer however if we go deeper with the file mode as well for Windows as it is not recommended to use them with Windows either.

DavidS:
Thanks for the work you have put in to this pull request. However the code in question is replaced by the native concat_file type, so this does not apply anymore.

PR #260, opened 2015-02-10
MODULES-1764 Fix missing method for check_is_owned_by for windows

cyberious:


cyberious:
Good point, I was going for easy to remember to rip out as well.

cyberious:
@mhaskel I moved into a separate file.  Had to add another as well for the case when files contain spaces.

PR #261, opened 2015-02-10
MODULES-1456 - make sure ruby is in path on PE

mhaskel:
Ruby is not in the path for all PE installations, so make sure the
concat script is run with ruby in the path.


raphink:
This fails with strict_variables because `is_pe` is not a standard fact.

PR #263, opened 2015-02-12
Fix validate_cmd file resource parameter

cmurphy:
The validate_cmd parameter is only valid on newer versions of puppet.
This module is supporting all 3.x versions of puppet. This commit only
applies the validate_cmd parameter to the file resource if it is
defined and adds documentation noting that it will not work on older
versons of puppet.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -185,8 +185,14 @@ define concat(
       path         => $path,
       alias        => "concat_${name}",
       source       => "${fragdir}/${concat_name}",
-      validate_cmd => $validate_cmd,
       backup       => $backup,
+    }
+
+    # Only newer versions of puppet 3.x support the validate_cmd parameter
+    if $validate_cmd {
+      File[$name] {
+        validate_cmd => $validate_cmd,
+      }
     }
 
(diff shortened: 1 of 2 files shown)


domcleal:
:+1: fixes the issue for me.

manifests/init.pp
```diff
@@ -185,10 +185,16 @@
...
-      validate_cmd => $validate_cmd,
       backup       => $backup,
     }
 
+    # Only newer versions of puppet 3.x support the validate_cmd parameter
+    if $validate_cmd {
+      File[$name] {
+        validate_cmd => $validate_cmd,
+      }
+    }
```
igalic: unless i'm entirely mistaken: given that `validate_cmd`'s default is `undef`, we shouldn't really need the `if` here.

domcleal: If the validate_cmd parameter is unconditionally added to the `file {}` resource above then pre-3.5 versions will throw an error, irrespective of whether it's undef or set.

igalic: ack. i did see the patch by @petems a bit too late before commenting here

PR #264, opened 2015-02-12
Fix validate_cmd usage in older Puppet

petems:
Puppet < 3.5 doesn't have validate_cmd on the file parameter. This allows us to add it 
to older puppet without having to restrict versions

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -196,7 +196,7 @@ define concat(
         }
       }
       else {
-        validate_cmd("${fragdir}/${concat_name}", $validate_cmd)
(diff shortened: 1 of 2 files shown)


ghoneycutt:
metadata.json needs to bump stdlib from 3.2.0 to 4.2.0 to match .fixtures.yml and because 3.2.0 is missing validate_cmd() which is introduced here.

ghoneycutt:
@petems awesome fix :)

petems:
@ghoneycutt Good catch, yeah the .fixtures spec file had it, forgot to check about the actual module!

petems:
Since we can't use stdlib's validate_cmd until it's released as an official version, I'm closing this for now

PR #266, opened 2015-02-13
Remove shell script

mhaskel:
The ruby script works just as well and works on all supported platforms. No need to maintain both.


igalic:
:+1: :100: for less code!

PR #272, opened 2015-02-19
Update fragment.pp

tpdownes:
The fragment code creates a bunch of intermediate files and this patch ensures that they are safely created regardless of the 'order' parameter given. Context: I am creating a grid certificate mapping file whose unique keys, on which the file should be sorted, are certificate subjects like this:

"/DC=org/DC=cilogon/C=US/O=LIGO/CN=Thomas Downes thomas.downes@ligo.org"

This code ensures that the files created are safe using this ordering scheme.

### DIFF
--- a/manifests/fragment.pp
+++ b/manifests/fragment.pp
@@ -68,6 +68,7 @@ define concat::fragment(
 
   $safe_name        = regsubst($name, '[/:\n]', '_', 'GM')
   $safe_target_name = regsubst($target, '[/:\n]', '_', 'GM')
+  $safe_order       = regsubst($order, '[/:\n]', '_', 'GM')
   $concatdir        = $concat::setup::concatdir
   $fragdir          = "${concatdir}/${safe_target_name}"
   $fragowner            = $concat::setup::fragment_owner
@@ -110,7 +111,7 @@ define concat::fragment(
 
   # punt on group ownership until some point in the distant future when $::gid
   # can be relied on to be present
-  file { "${fragdir}/fragments/${order}_${safe_name}":
+  file { "${fragdir}/fragments/${safe_order}_${safe_name}":
     ensure  => $safe_ensure,
     owner   => $fragowner,
     mode    => $fragmode,


manifests/fragment.pp
```diff
@@ -68,6 +68,7 @@
 
   $safe_name        = regsubst($name, '[/:\n]', '_', 'GM')
   $safe_target_name = regsubst($target, '[/:\n]', '_', 'GM')
+  $safe_order       = regsubst($order, '[/:\n]', '_', 'GM')
```
igalic: will this work when order is numeric?

tpdownes: Depends on what you mean by work.

According to the documentation, the order parameter is a string regardless of whether sorting is alpha or numeric. The function will replace the 3 characters in the regexp with underscores. None of those 3 characters are valid in a number. So, in some sense, this will let you replace a broken non-number with a different broken non-number that happens to be more filesystem-safe.

I have to admit it's not clear to me why the module differentiates between alpha sorting and numerical sorting when they are both instances of lexical sorting.

mhaskel: @tpdownes the code behaves differently depending on the sorting type [here](https://github.com/puppetlabs/puppetlabs-concat/blob/master/files/concatfragments.rb#L124-L141)

I think my main concern in this is around the use of the underscore, in the case where the ordering is set to numeric and there are invalid characters in the order, it will only sort on the order up to the first invalid character. Could you update the `regsubst` to replace it with a dash maybe instead?

tpdownes:
Truth be told, you should do with the pull request as you wish. I don't see how the characters I'm replacing are any more valid numerically than the underscore and therefore I conclude that what you mean by numeric sorting is something different from what I would guess.

mhaskel:
@tpdownes thanks for the contribution, and thanks for the feedback. We decided to go with validating the `$order` string (#275), so I'm going to close this PR. Thanks again!

PR #273, opened 2015-02-23
Lookup is_pe fact with getvar

cmurphy:
If the system is not running PE, the $::is_pe fact will be undefined,
causing errors when run under strict variables mode. This patch uses
the getvar() function to look up the fact so that it won't fail if not
defined.

Alternative to https://github.com/puppetlabs/puppetlabs-concat/pull/270

This should fix the broken CI for the postgresql module.


hunner:
`getvar` added in stdlib 0.1.7, so we should be good there :)

PR #275, opened 2015-02-25
Add validation for order parameter

mhaskel:
Per discussion on #272, we're going to validate $order rather than worry
about needing to make substitutions on it.


hunner:
Closes #272

PR #276, opened 2015-02-26
Fixup the usage of the ruby interpreter for OpenBSD to run

buzzdeee:
concatfragments.rb, similar to what's done on Windows.

OpenBSD "breakage" was introduced with Git commit:
7eb14b776838039ac923e8ff8e56871a3b35df1a

The ruby interpreter used by Puppet package for OpenBSD is 
called ruby21, therefore the shebang in concatfragments.rb
doesn't work, and breaks for me on at least OpenBSD 5.6 and 5.7-beta.

Whenever the default ruby interpreter used by Puppet on OpenBSD will be 
updated, then a version check for ::kernelversion or the like 
can/must be added.

please consider inclusion of the fix, or any feedback appreciated.

cheers,
Sebastian

### DIFF
--- a/manifests/setup.pp
+++ b/manifests/setup.pp
@@ -37,6 +37,8 @@ class concat::setup {
 
   $script_owner = $::osfamily ? { 'windows' => undef, default => $::id }
 
+  $script_group = $script_owner ? { 'root' => '0', default => undef }
+
   $script_mode = $::osfamily ? { 'windows' => undef, default => '0755' }
 
   $script_command = $::osfamily? {
@@ -51,6 +53,7 @@ class concat::setup {
   file { $script_path:
     ensure => file,
     owner  => $script_owner,
+    group  => $script_group,
     mode   => $script_mode,
     source => "puppet:///modules/concat/${script_name}",
   }


jhoblitt:
Ugh - the ports package for ruby does not install a symlink/wrapper for 'ruby'? This needs unit test coverage before it can be merged.

buzzdeee:
nope, it (un)fortunately doesn't. 

One reason is, because you can install multiple versions of ports/packages of ruby in parallel without conflicting. Then you decide as administrator, or user, what version you want as default, by either setting global or personal symlinks to the versioned binaries in question.

The very same is true for pyhton, the binary is installed as python2.7 for example. 

It is just, within the ports infrastructure, it happens that ruby 2.1 is the "defined" default, and that's also the what puppet uses. I.e. the puppet-dashboard port/package explicitly requires the ruby 1.8 as default, and it lives quite happily with puppet next to each other.

Another example is the gem package provider that comes with Puppet. There is a patch on the port, to explicitly use 'gem21', instead of just 'gem' to install ruby gems via Puppet.

I'll see if I can get a test case written for it over the weekend, guess/hope there is one for Windows, that I can just adapt ;)

cheers,
Sebastian

buzzdeee:
I was looking for a specific case for Windows, but didn't found one. I'm not that good at writing tests (yet), and I don't really know, how to write a test for OpenBSD 
However, I added a note to the README.md that, whenever the Ruby version used by Puppet changes, this must be enhanced to check for the OpenBSD version. Is that sufficient?

Otherwise, this file could also be converted into a template, and the ruby interpreter, defaulting to what it currently is, be made a variable.

buzzdeee:
rebased...

PR #278, opened 2015-03-02
Add show_diff to concat/concat::fragment resources.

diranged:
Allows keeping our concat diffs private during puppet runs so that
secrets are not logged.

### DIFF
--- a/manifests/fragment.pp
+++ b/manifests/fragment.pp
@@ -23,20 +23,24 @@
 #   Deprecated
 # [*backup*]
 #   Deprecated
+# [*show_diff*]
+#   Boolean whether to show fragment-diffs or not.
 #
 define concat::fragment(
     $target,
-    $content = undef,
-    $source  = undef,
-    $order   = '10',
-    $ensure  = undef,
-    $mode    = undef,
-    $owner   = undef,
-    $group   = undef,
-    $backup  = undef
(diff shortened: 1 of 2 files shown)


manifests/fragment.pp
```diff
@@ -23,20 +23,24 @@
...
-    $backup  = undef
+    $content   = undef,
+    $source    = undef,
+    $order     = '10',
+    $ensure    = undef,
+    $mode      = undef,
+    $owner     = undef,
+    $group     = undef,
+    $backup    = undef,
+    $show_diff = true
```
igalic: why is this true by default?
shouldn't we just pass undef down the line to `file`, so it uses `file`'s default?

igalic:
general comments:

this pull request now needs a [rebase](http://git-scm.com/book/en/Git-Branching-Rebasing). it also needs Documentation and spec tests.

kincl:
Is there any movement on this? Should I create a new pull request to add this functionality or wait for this one?

bmjen:
@kincl Thanks for the contribution and apologies for the lack of response. Can you please reopen this PR against either 1.2.x or 2.0.x, whichever release stream this feature should belong to. We were in the midst of refactoring the concat module, so the PRs got shelved a bit.

Thanks.

PR #279, opened 2015-03-05
Set script's group to root if script owner is root

thias:
When files on a puppetmaster are owned by a non-root user, concatfragments.sh
gets installed on the nodes with a group owner matching the one of the master.

This has no security implications since the file is mode 755, but does lead to
possible ping-pong situations when switching between environments where the
files on the master have different group ownership.

### DIFF
--- a/manifests/fragment.pp
+++ b/manifests/fragment.pp
@@ -42,7 +42,7 @@ define concat::fragment(
   }
   if !(is_string($order) or is_integer($order)) {
     fail('$order is not a string or integer.')
-  } elsif $order =~ /[:\n\/]/ {
+  } elsif (is_string($order) and $order =~ /[:\n\/]/) {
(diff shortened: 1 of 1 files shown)


manifests/setup.pp
```diff
@@ -37,6 +37,8 @@
 
   $script_owner = $::osfamily ? { 'windows' => undef, default => $::id }
 
+  $script_group = $script_owner ? { 'root' => 'root', default => undef }
+
```
igalic: this will not work on some BSDs…

thias: Darn, you're certainly right. Do you have any idea how to solve my problem, then? It's annoying to have puppet default to the remote file owner+group instead of the local user and his default group...

igalic: use `0` instead of `'root'`

PR #280, opened 2015-03-05
Set script's group to 0 if script owner is root

thias:
When files on a puppetmaster are owned by a non-root user,
concatfragments.sh gets installed on the nodes with a group
owner matching the one of the master.

This has no security implications since the file is mode 755,
but does lead to possible ping-pong situations when switching
between environments where the files on the master have
different group ownership.

Use '0' instead of 'root', since the root user's main group
isn't always 'root' on some BSDs, but always '0'.

### DIFF
--- a/manifests/setup.pp
+++ b/manifests/setup.pp
@@ -37,6 +37,8 @@ class concat::setup {
 
   $script_owner = $::osfamily ? { 'windows' => undef, default => $::id }
 
+  $script_group = $script_owner ? { 'root' => '0', default => undef }
+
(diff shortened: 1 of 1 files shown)


sodabrew:
I'm seeing this problem now that I've updated my concat module to 1.2.0 - the group owner of `/var/lib/puppet/concat/bin/concatfragments.sh` is all over the place - ntp, postfix, root, haproxy - looks like whatever was the last value set for any other file in the catalog just prior to this file!

hunner:
:+1:

daenney:
:+1:

sodabrew:
:trophy:

PR #284, opened 2015-03-26
Fix breakage on OpenBSD in similar fashion as it is done for Windows.

buzzdeee:
Broken since the switch from the shell script to the ruby script.

I did something stupid with git while trying to rebase, this is the same as what I had in #276, fixing the breakage introduced with the switch from concatfragments.sh to concatfragments.rb

### DIFF
--- a/manifests/setup.pp
+++ b/manifests/setup.pp
@@ -43,6 +43,7 @@ class concat::setup {
 
   $script_command = $::osfamily? {
     'windows' => "ruby.exe '${script_path}'",
+    'openbsd' => "/usr/local/bin/ruby21 '${script_path}'",
     default   => $script_path
   }
 


jhoblitt:
It really feels like the ruby binary path should be coming from a fact...

Could you add unit test coverage for this change?

buzzdeee:
I can also remove the /usr/local/bin, and just have it ruby21. It is in the default PATH anyways, so don't know, if a fact is really required?
Also, the path where to find the ruby binary, if at all, should likely come from another module, i.e. puppetlabs-ruby? At least I think that would be more appropriate. But then, this module would need a dependency to puppetlabs-ruby, at least for OpenBSD. Would this be acceptable, worthwile?

On the other hand, I could add a fact, similarily like done to the puppetlabs-java module:
https://github.com/puppetlabs/puppetlabs-java/pull/109

Let me know what you prefer, and I'll look into that.

Unit test, I'll see what I can do, I'm not an expert in that one (yet ;)



cheers,
Sebastian

jhoblitt:
I wouldn't object to a dependency on `puppetlabs-ruby` since this module now executes a local ruby script on all platforms.  However, that module doesn't at present have a fact for the ruby interpreter path or support for OpenBSD.

Adding that support is the direction I would be inclined to go since, as far as I can tell, [ports](
http://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/sysutils/ruby-puppet/Makefile.inc?rev=1.29&content-type=text/x-cvsweb-markup) doesn't mandate a ruby version.

It would be good to get input from the PL folks before we change the dep chain.

buzzdeee:
Hrm, I feared you'd say that depending on puppetlabs-ruby would be an option ;)
anyways, you are right that the Puppet port itself doesn't specify a ruby version to build against, however, that is managed in the Ruby port module, see:

http://cvsweb.openbsd.org/cgi-bin/cvsweb/~checkout~/ports/lang/ruby/ruby.port.mk?rev=1.79&content-type=text/plain
where it defines the default FLAVOR:

FLAVOR?=
# Without a FLAVOR, assume the use of ruby 2.1.
.    if empty(FLAVOR)
FLAVOR =                ruby21
.    endif

Since the Puppet port doesn't define a FLAVOR, its building/depending on ruby 2.1.x per default. 
This is true for OpenBSD 5.6, and 5.7. Not sure if 5.8 will have Ruby 2.2.x as default.

Something that is even "worse", is that all ruby versions, can be installed in parallel, without issues, and just works. So just enumerating, if ruby18, ruby19, ruby20, ruby21, etc. is available or not, is not sufficient to figure out, which Ruby version here for the concat module. The only way would be to figure out the installed puppet package version, i.e. 
# pkg_info puppet|grep Information for
Information for inst:puppet-3.7.4p1
# cat /var/db/pkg/puppet-3.7.4p1/+REQUIRING,
ruby21-augeas-0.5.0p3
ruby21-shadow-2.4.1
gtar-1.28p0
ruby-2.1.5
ruby21-hiera-1.3.4p1
ruby21-json-1.8.1
ruby21-rgen-0.7.0
facter-2.3.0

and from that output, deduce what ruby version to use.
as far as I can see, pkg_info only has an option to show what packages depend on a given package, but doesn't show, what are direct dependencies of that queried package.

This is to say, a lot to implement, I think it would be easier, instead of adding the full path, I'd only stick with the interpreter name, i.e. ruby21. /usr/local/bin is in the default PATH.
If additionally wanted, I could match on kernelversion or kernelmajorversion to restrict usage of ruby21 only to OpenBSD 5.6, and 5.7. I don't remember correctly, whether OpenBSD 5.5 had 1.8.X as default Ruby version, or whether it was Ruby 2.1 already. The switch to 2.1 is about a year ago, so 5.5 may or may not had it already.

PR #285, opened 2015-03-29
setup: set user/group explicitly for dirs & script

j-vizcaino:
Hi,

I think user and group properties need to be set explicitly for the concat *runtime environment* (base dir and script).
As Puppet propagates resources defaults to included modules, declaring something like

```puppet
File {
  owner => bind,
  group => root
}
```
in a module makes concat create its basedir belonging to user `bind`.

Jérôme

### DIFF
--- a/manifests/setup.pp
+++ b/manifests/setup.pp
@@ -35,9 +35,9 @@ class concat::setup {
 
   $script_path = "${concatdir}/bin/${script_name}"
 
-  $script_owner = $::osfamily ? { 'windows' => undef, default => $::id }
+  $default_owner = $::osfamily ? { 'windows' => undef, default => $::id }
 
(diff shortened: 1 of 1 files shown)


jhoblitt:
This seems reasonable but needs unit test coverage.

j-vizcaino:
I think this is already tested in spec/acceptance/concat_spec.rb:44

jhoblitt:
This PR adds new parameters to existing file resources.  https://github.com/puppetlabs/puppetlabs-concat/blob/master/spec/unit/classes/concat_setup_spec.rb#L26-L34

j-vizcaino:
Ok, thanks for pointing the correct unit test.
I'll try to implement the checks for proper user/group ownership.

j-vizcaino:
Closing this in order to isolate the commits in a proper branch and open another PR

PR #288, opened 2015-04-06
(MODULES-1700) Change the filebucketing behavior so static_compiler can ...

woneill:
...handle them

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -141,7 +141,7 @@ define concat(
   }
 
   File {
-    backup  => false,
+    backup  => $backup,
   }
 
   # reset poisoned Exec defaults
--- a/manifests/setup.pp
+++ b/manifests/setup.pp
@@ -47,10 +47,6 @@ class concat::setup {
     default   => $script_path
   }
 
-  File {
-    backup => false,
-  }
-
   file { $script_path:
     ensure => file,
     owner  => $default_owner,
--- a/manifests/fragment.pp
+++ b/manifests/fragment.pp
@@ -57,6 +57,7 @@ define concat::fragment(
   if $backup {
     warning('The $backup parameter to concat::fragment is deprecated and has no effect')
   }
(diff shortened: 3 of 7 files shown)


manifests/init.pp
```diff
@@ -141,7 +141,7 @@
   }
 
   File {
-    backup  => false,
+    backup  => $backup,
```
hunner: I don't think this is needed to work with the static compiler. None of these files have sources pointing at the puppet fileserver.

manifests/setup.pp
```diff
@@ -46,10 +46,6 @@
     default   => $script_path
   }
 
-  File {
-    backup => false,
-  }
```
hunner: Instead of changing the default, could you just change the backup value for the `File[$script_path]` resource?

jhoblitt: Resource defaulting has caused so many problematic side effects I'm happy to see it go...

hunner:
I merged some things and now it needs a rebase, also.

manifests/init.pp
woneill: I tried reverting just this line and it looks like concat::setup inherited it:
```
Error: /Stage[main]/Concat::Setup/File[/var/lib/puppet/concat/bin/concatfragments.rb]/ensure: change from absent to file failed: Could not retrieve content for {md5}b684db0eac243553a6a79365119a363d from filebucket
```

hunner: Okay, it's fine then.

manifests/setup.pp
woneill: I couldn't come up with a way to get the value of $backup inside concat::setup short of making it a parameterized class. Plus it seems that this is being inherited anyway

woneill:
I'll handle the rebase and update

woneill:
I rebased then did a `git push --force`

PR #304, opened 2015-04-24
update metadata.json with correct dependencies and setting exact version...

bmjen:
... for file_concat dependency


mhaskel:
:+1:

PR #307, opened 2015-05-01
readd ensure_newline param and tests for backwards compatibility

bmjen:


lib/puppet/type/concat_file.rb
```diff
@@ -134,6 +140,11 @@ def fragment_content(r)
       tmp = Puppet::FileServing::Content.indirection.find(@source, :environment => catalog.environment)
       fragment_content = tmp.content unless tmp.nil?
     end
+
+    if self[:ensure_newline]
+      fragment_content<<"\n"
```
cyberious: I think we should be checking to see if it has a newline before appending a newline.  If it already ends in \n we will end up making a double newline.

cyberious: Something like
```
fragment_content << "\n" unless fragment_content =~ /\n$/
```

PR #308, opened 2015-05-01
(maint) Only add newline if string does not end with newline

cyberious:


bmjen:
I like this change, however.. I did it the other way to maintain consistency with how the old concat did it :-\

https://github.com/puppetlabs/puppetlabs-concat/blob/1.2.x/files/concatfragments.rb#L135-L137

cyberious:
I would argue that it is a bug than and we should not be adding an additional newline

PR #309, opened 2015-05-03
(maint) fixup invalid JSON

justinstoller:
Previously there was a trailing comma after the last item in a list.
That is invalid in JSON. This commit removes that one comma.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -105,6 +105,6 @@
     }
   ],
   "dependencies": [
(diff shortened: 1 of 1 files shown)


justinstoller:
This is often helpful when working directly with JSON: http://jsonlint.com/

igalic:
there's also @puppet-community's metadata-json-lint, which has become a quasi-community standard.

PR #317, opened 2015-05-27
MODULES-2054 - fixes dependency bug in creating the target file

bmjen:


spec/acceptance/fragment_source_spec.rb
```diff
@@ -149,7 +149,6 @@
     end
     describe file("#{basedir}/fail_no_source") do
       #FIXME: Serverspec::Type::File doesn't support exists? for some reason. so... hack.
-      it { should_not be_file }
```
mhaskel: why is this being removed? Shouldn't things that are not files still not be files?

bmjen: Had a convo with @hunner about this. Since we've moved the resource creation to `generate` and updating of the content to `eval_generate`, if we fail to fetch the content from the fragments and the file doesn't exist, the apply will create an empty file. If the file does exist, it will be left unmodified. This is different from previous behavior but I think this is an edge case and the impact that this change could cause doesn't outweigh the complexity we'd have to add to the `eval_generate` function in order to properly handle cleaning up the resource in the catalog on rescue. 

¯\_(ツ)_/¯

mhaskel: :+1:

spec/acceptance/concurrency_test.rb
```diff
@@ -0,0 +1,37 @@
...
+        recurse => true,
+      }
+
+      concat { "foobar":
+        ensure => 'present',
+        path   => '#{basedir}/bar/foobar',
+      }
+
+      concat::fragment { 'foo':
+        target => '#{basedir}/bar/foobar',
```
mhaskel: I think this needs to be `target => 'foobar'`

PR #318, opened 2015-05-28
fix fragment target handling

bmjen:
Needed to bind fragment targets to it's parent concat resource's title so that we can
find them in the catalog. Throws a warning if the target is not found.

### DIFF
--- a/lib/puppet/type/concat_fragment.rb
+++ b/lib/puppet/type/concat_fragment.rb
@@ -17,6 +17,10 @@ Puppet::Type.newtype(:concat_fragment) do
     desc "Unique name"
   end
 
+  newparam(:target) do
+    desc "Target"
+  end
+
   newparam(:content) do
     desc "Content"
   end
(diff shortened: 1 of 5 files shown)


lib/puppet/type/concat_fragment.rb
```diff
@@ -38,11 +42,22 @@
...
+    # Check if target is set
+    fail Puppet::ParseError, "Target not set" if self[:target].nil?
+
     # Check if either source or content is set. raise error if none is set
     fail Puppet::ParseError, "Set either 'source' or 'content'" if self[:source].nil? && self[:content].nil?
 
     # Check if both are set, if so rais error
     fail Puppet::ParseError, "Can't use 'source' and 'content' at the same time" if !self[:source].nil? && !self[:content].nil?
+
+    self[:tag] = regsubst(self[:target], '[/:\n\s]', '_', 'GM') if self[:tag].nil?
```
mhaskel: doesn't this change the required-ness of tag? or at least change the default? May be a feature, should have docs, should maybe be split out.

bmjen: You're right. forgot to remove that bit. Need to add a check in the validation block too.

PR #319, opened 2015-05-28
adds file autorequire

bmjen:
autorequire will add parent directories of the concat_file to the dependency chain.
So they can be created in the right order if they do not exist.


cyberious:
:+1:

3flex:
Thanks for this. You can close MODULES-2023 as appropriate (I'd raised this issue there).

PR #320, opened 2015-05-29
(MODULES-2080) Call out changed behaviour of 'warn' parameter

DavidS:
Before 2.0.0, this parameter would add a newline at the end of the warn
message. To improve flexibilty, this was removed. Please add it
explicitely if you need it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,8 +183,12 @@ Specifies whether to add a header message at the top of the destination file. Va
 If you set 'warn' to 'true', `concat` adds the following message:
 
 ~~~
-# This file is managed by Puppet. DO NOT EDIT.
+# This file is managed by Puppet. DO NOT EDIT.\n
 ~~~
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -183,9 +183,13 @@ Specifies whether to add a header message at the top of the destination file. Va
 If you set 'warn' to 'true', `concat` adds the following message:
 
 ~~~
-# This file is managed by Puppet. DO NOT EDIT.
+# This file is managed by Puppet. DO NOT EDIT.\n
```
igalic: not really sure how useful the \n here is in the README…

DavidS: I wanted to note the fact that this string is copied verbatim to the file. I could leave off the `\n` and instead s/following message/following line/ in the paragraph above?

igalic: *nod*

PR #327, opened 2015-06-03
(MODULES-2094) Extend regexp to remove parenthesis on safe names

luisfdez:
This patch allows concatenation on folders like `C:\Program Files (x86)` on Windows agents.


bmjen:
Hi @luisfdez I PR'd on top of your fix and added an acceptance test. Thanks for the fix and great find!

PR #337, opened 2015-06-11
(MODULES-2104) Workaround for PUP-1963

reidmv:
This is kind of a brute-force workaround for PUP-1963. I'm putting it up for discussion in case PUP-1963 is punted until Puppet 5.x and some way of working around the issue has to be found.

PUP-1963 captures a bug in which generated resources do not receive dependency edges. Because concat_file works by generating a file resource, until PUP-1963 is resolved notify and subscribe relationships against a concat or concat_file resource will not work properly.

This commit works around the issue by causing the concat_file resource itself to emit the event.

The generated file resource captures the desired state of the file. Ideally, the concat_file resource could remain an effective "metaresource" and simply pass through events, but that's currently not possible. Therefore when the concat_file resource is evaluated this commit causes it to "look ahead" to see whether or not the generated file resource is insync. If the generated file resource is not insync, the concat_file resource itself will declare that it is not insync, and so emit the correct event. Resources which notify or subscribe to the concat_file resource will correctly receive the emitted event.

The downside to this workaround is an extra event log any time the concat_file resource is updated.

### DIFF
--- a/lib/puppet/type/concat_file.rb
+++ b/lib/puppet/type/concat_file.rb
@@ -24,6 +24,45 @@ Puppet::Type.newtype(:concat_file) do
     defaultvalues
 
     defaultto { :present }
+
+    # Workaround for PUP-1963 (https://tickets.puppetlabs.com/browse/PUP-1963)
+    #
+    # Because notify/subscribe relationships are not propogated through to
+    # generated resources, the insync status of the concat_file resource is
+    # configured to be whether or not the generated file resource is insync.
+    # This will result in subscribe relationships working correctly.
+    def insync?(is)
+      # Determine if there are changes pending to the generated file resource
+      generated_file = resource.catalog.resource("File[#{@resource[:path]}]")
+      current_values = generated_file.retrieve_resource.to_hash
+      pending_change = generated_file.properties.any? do |property|
+        current_value = current_values[property.name]
+        property.should && !property.safe_insync?(current_value) ? true : false
+      end
+
+      # If no changes are pending, then the concat resource is in sync
+      !pending_change
+    end
+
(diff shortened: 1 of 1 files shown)


reidmv:
Upon further reflection this would still need more work to become fully functional. Events are emitted when appropriate but the generated file resource is not contained by the concat_file resource so there's no guarantee that the generated resource itself won't be evaluated *after* resources subscribed the the concat_file.

hunner:
@reidmv If I understand how puppet works, the concat -> file dependency is more like a "concat contains file" thing... The file resource ALWAYS happens immediately after the concat resource, so any resources that the concat refreshes should always happen after the file resource is evaluated.

reidmv:
@hunner that was how I assumed it would work too but I haven't been able to validate it. When I play around with `--ordering=random`, I occasionally get results that aren't dependency correct. The concat_file resource itself will happen before a subscribed resource, but the generated file might happen afterwards (unfortunately). At least with only this change.

hunner:
Okay, I'll look into that.

My PR takes a different approach and copies the dependencies from the parent concat_file resource to the generated file resource, though @nibalizer gave a better idea for how to do it.

reidmv:
I like the idea of actually building in the correct dependencies a lot better. :+1:

hunner:
Except that it requires mucking with the graph directly, which is kind of iffy. Not sure it's considered a public API.

nibalizer:
Devils advocate: Puppet 3 isn't making any more significant changes so all apis (public or private) will not change.

nibalizer:
Then the proper puppet fix covers the Puppet 4 case

reidmv:
Closing in favor of #338

PR #338, opened 2015-06-11
(MODULES-2104) Update graph from generate method

hunner:
This is a workaround for PUP-1963. It involves getting the refresh edges (called `builddepends`) from the parent `concat` resource and adding them to the generated `file` resource.

One requirement is to rename the `concat_file` type to `concat` to get the builddepends to show up; for some reason trying to get builddepends of a define does not work.

TODO: The type should add the warn_header to the top of the should_content, because the concat define used to do this but is now gone.


hunner:
Different implementation from #337

reidmv:
Getting the builddepends of the define might not work because defines are implemented like classes, with Whits. If that's the case, it'd probably be possible to retrieve the builddepends from the relavant Whit(s).

Personally I'm all in favor of eliminating the defined type anyway.

reidmv:
I prefer this implementation over #337.

lollipopman:
This worked well for me after making this change to the concat type, so that the tag defaults to concat name.
```
newparam(:tag) do
  desc "Tag reference to collect all concat_fragment's with the same tag"
  defaultto do
    resource.value(:name).gsub(/[\/:\n\s\(\)]/, '_')
  end
end
```

lollipopman:
This change was also necessary @hunner
```
    concat: when adding dependency edges check which side we are on

    Determine if we are the source or the target of the dependency edge and
    modify accordingly, this also avoids creating dependency cycles

diff --git a/vendor/modules/concat/lib/puppet/type/concat.rb b/vendor/modules/concat/lib/puppet/type/concat.rb
index ca038b1..8a33588 100644
--- a/vendor/modules/concat/lib/puppet/type/concat.rb
+++ b/vendor/modules/concat/lib/puppet/type/concat.rb
@@ -173,8 +173,13 @@ Puppet::Type.newtype(:concat) do
     deps = catalog.relationship_graph.dependents(self).map(&:builddepends)
     [deps,self.builddepends].flatten.each do |edge|
       newedge = edge.dup
-      newedge.source = catalog.resource("File[#{self[:path]}]")
-      catalog.relationship_graph.add_edge(newedge)
+      if newedge.source == self
+        newedge.source = catalog.resource("File[#{self[:path]}]")
+        catalog.relationship_graph.add_edge(newedge)
+      elsif newedge.target == self
+        newedge.target = catalog.resource("File[#{self[:path]}]")
+        catalog.relationship_graph.add_edge(newedge)
+      end
     end
     []
   end
```

reidmv:
@hunner @lollipopman I've formatted the changes suggested as a PR to Hunner's branch. Hunner merging that in to his fork should result in the changes being incorporated into this PR. See https://github.com/hunner/puppetlabs-concat/pull/2.

reidmv:
I took a stab at starting to convert the old defined-type tests to apply to the new native type.

https://github.com/reidmv/puppetlabs-concat/blob/rename_init_tests/spec/unit/puppet/type/concat_spec.rb

If that's moving in the right direction I can finish it off.

lollipopman:
@reidmv thanks, I think this look good, can we also eliminate the concat::fragment define type?

reidmv:
@lollipopman I agree that there's no longer any direct value in having a concat::fragment defined type, and it would be awesome to remove it just to eliminate a redundant vertex from the graph. That's a little tricker than concat though since eliminating concat::fragment would change the API. If we leave it there, the 2.x release will be a mostly drop-in replacement for people who were previously using the 1.x concat.

I think the concat::fragment defined type should be trimmed down to be just a simple wrapper around concat_fragment, but that's probably work for a different ticket.

lollipopman:
@reidmv Given that this will be a 2.0 release, I think it would be the right time to get rid of the define type, I think it adds unnecessary confusion when you discover that there is concat_fragment type used by the define type. Also the new version requires other changes such as changing all your dependencies from File["/foo/bar.txt"] to Concat["/foo/bar.txt"]

bmjen:
@hunner We should re-PR this against the 2.0.x branch. Since we reverted master to 1.2.x state.

Also, since we released 2.x with the defines `concat` and `concat::fragment` and types `concat_file` and `concat_fragment`, I'm thinking this change will result in a 3.0 release.

PR #341, opened 2015-06-13
Associate fragments by path, not tag

reidmv:
Previously tag was declared as a parameter for concat_file and also concat_fragment (also the name of a metaparameter), and the tag value was used to associate concat_fragment resources with their concat_file collector.

ANAICT this was an unnecessarily brittle means of associating collections. Basing association on tags effectively relied on the end user not using the tag parameter, which is not a safe assumption given that "tag" is also a useful metaparameter.

This commit makes path the namevar of the concat_file type and changes the association to be between the concat_file-path and
concat_fragment-target parameters.

### DIFF
--- a/lib/puppet/type/concat_file.rb
+++ b/lib/puppet/type/concat_file.rb
@@ -30,15 +30,7 @@ Puppet::Type.newtype(:concat_file) do
     self[:ensure] == :present
   end
 
-  newparam(:name, :namevar => true) do
-    desc "Resource name"
-  end
-
-  newparam(:tag) do
-    desc "Tag reference to collect all concat_fragment's with the same tag"
-  end
-
-  newparam(:path) do
+  newparam(:path, :namevar => true) do
     desc "The output file"
     defaultto do
       resource.value(:name)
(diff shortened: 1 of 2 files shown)


reidmv:
I noticed this behavior while thinking about the impact of @lollipopman's suggested default tag parameter value in #338. I'm curious if there's any kind of original use case or rationale that is enabled when tags are used instead of path/target, but even if there is I'd be concerned about overloading "tag". Thoughts?

lollipopman:
@reidmv I think this change makes sense, I cannot think of a good use case where the user would want to manipulate the association themselves.

bmjen:
@reidmv Just some history. When @morgan and I made the change to fix a bug in properly associating concat_fragments to the concat_file title (and not path), we left the tag param there because removing it would result in an undesirable symver update. :(

However, I'm not sure if that's a valid concern anymore...

reidmv:
@bmjen thanks for the context, that makes sense.

If we transition away from concat being a defined type which wraps and carefully controls tag on the nested concat_file resource, then this issue becomes much more important. Of course, tag will still be a valid parameter even if we remove it from the type, it'll just be the metaparameter tag and not the type-specific parameter tag...

I don't really know what happens when a type declares a parameter that conflicts with an existing metaparameter. That ambiguity alone is enough to make me want to steer clear of it.

reidmv:
Is there anything blocking a decision on this PR?

morgan:
@bmjen Wrong Morgan :)

bmjen:
My apologies @morgan!

bmjen:
@reidmv I'm going to close this as the 2.0.x has been put on the backburner until the puppet bug PUP-1963 is released. At that point, we'll do some further refactoring of the type that should alleviate this problem.

PR #342, opened 2015-06-16
(MODULES-2104) Update graph from generate method

hunner:
This is a retarget of #338 to the 2.0.x branch with suggested changes incorporated.

> This is a workaround for PUP-1963. It involves getting the refresh edges (called builddepends) from the parent concat resource and adding them to the generated file resource.
>
> One requirement is to rename the concat_file type to concat to get the builddepends to show up; for some reason trying to get builddepends of a define does not work.
>
> TODO: The type should add the warn_header to the top of the should_content, because the concat define used to do this but is now gone.


reidmv:
@bmjen or other maintainers: is there anything else blocking a decision on this PR?

hunner:
@reidmv There was a question as to whether this would need a 3.0.0 because of removing the concat_file resource (renaming, really). I'd rather the answer be "lolz no; no one had a chance to even explicitly use that" and move forward with 2.0.2-rc1 (because yeah, these will only be rc releases for a bit).

reidmv:
If it's a serious concern we could just drop a concat_file wrapper defined type in too. I'm also in favor of 2.x, but not strongly. I'd be just as happy skipping straight to 3.x. Main thing I want to see is a merge & tag.

Who makes the version call? Would it be helpful for me to add in that defined type?

bmjen:
Going to close this PR for now, as the workaround we provided here is undesirable. We've pulled the 2.0.x releases from Forge and are waiting for the PUP-1963 fix to be released. We'll clean up the 2.0.x branch at that point and re-introduce concat as a native type.

PR #336, opened 2015-06-10
Compare $::is_pe as bool, not string

raphink:
stdlib now provides an $::is_pe fact, which defaults to "false".
When compared as a string, it is evaluated to true.

PR #343, opened 2015-06-19
Compare $::is_pe as bool, not string

raphink:
stdlib now provides an $::is_pe fact, which defaults to "false".
When compared as a string, it is evaluated to true.

This is https://github.com/puppetlabs/puppetlabs-concat/pull/336 rebased on `master`


mhaskel:
Great, thanks @raphink !

PR #344, opened 2015-06-20
Add support for Solaris 12

drewfisher314:
Fairly self-explanatory. This simply adds Solaris 12 to metadata.json.


jonnytpuppet:
It would be appreciated if you could update the README for this module to indicated that Solaris 12 is compatible, but not supported by Puppet Labs (at this stage).

Thanks

PR #348, opened 2015-07-10
(#2208) FreeBSD: call /usr/local/bin/ruby explicitly

gwollman:
On FreeBSD systems, the Ruby interpreter is normally installed in
/usr/local/bin, which is not on the default PATH when running init
scripts.  Most users will have it in their PATH when logged in, which
causes concat to fail when run from init (or "service puppet start")
but not when run from directly the command line ("sudo puppet agent").
This follows the same strategy as OpenBSD, but avoid specifying the
Ruby interpreter version explicitly, because FreeBSD supports three
different versions of Ruby (with the default always being installed as
/usr/local/bin/ruby).


mhaskel:
Thanks @gwollman !

PR #340, opened 2015-06-12
(╯°□°)╯︵┻━┻ the 2.0.x rewrite

bmjen:

PR #353, opened 2015-07-19
Fixing tests for puppet < 3.5

mwhahaha:
This change addresses MODULES-2115 and is a backport of the fixes that
were included as part of #340.


tphoney:
Thanks for putting in the effort @mwhahaha , the unit tests are still failing.

mwhahaha:
@tphoney, fixed it. The cut & paste from #340 had a different format for the error detection. I have also verified that tests no longer fail under PUPPET_GEM_VERSION="~> 3.4.0" which is what we were hitting.

mwhahaha:
guess this got addressed in #354, closing...

PR #360, opened 2015-07-30
mergeback 1.2.x to master

jhoblitt:


mhaskel:
See @bmjen 's comment on https://github.com/puppetlabs/puppetlabs-concat/pull/359

jhoblitt:
@mhaskel Would it be better to revert #359 and then merge it master? (this is my fault)

bmjen:
@jhoblitt Is there a better way to solve #359 that won't break both usecases?

jhoblitt:
@bmjen There's several issues here.

* The release branch is ahead of master (it should have been merged to master at the last release)
* I merged a PR to the release branch when I shouldn't have
* There's the issue with the static compiler that I had forgotten about (it looks like I even commented on #288)

I'm going to go ahead and revert #359 from the release branch and merge everything else from the release branch into master.

jhoblitt:
I did a manual merge to master to exclude the merge/revert commits of #359 from landing on master.

jhoblitt:
@bmjen I don't use the static compiler but my impression is that the real issue here is resource defaults, is the correct?  I.e., if we just remove the resource defaults this becomes a non-issue?

PR #365, opened 2015-09-02
MODULES-1678 - Add show_diff attribute to concat defined type

jdkindy:
Add show_diff to concat and concat::fragment types.


tphoney:
@jdkindy This is a fantastic piece of work, i have a few comments though. Sorry.
Travis CI is failing and would need to be fixed.
There is no explicit test to test enabling diff and disabling the diff. I dont want the code to get broken accidentally.
Could you please squash your commits into a single commit, that describes the feature you have added.

jdkindy:
I will work on this - have never written the tests though so it may take a
little while.

PR #368, opened 2015-09-18
MODULES-1678 - Add show_diff attribute to concat and concat::fragment defined types

jdkindy:
Add show_diff to concat and concat::fragment types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,10 @@ Specifies a destination file for the combined fragments. Valid options: a string
 
 Specifies whether to overwrite the destination file if it already exists. Valid options: 'true' and 'false'. Default value: 'true'.
 
+#####`show_diff`
+
+Specifies whether to set the show_diff parameter for the file resource. Useful for hiding secrets stored in hiera from insecure reporting methods. Valid options: 'true
+
 #####`validate_cmd`
 
 Specifies a validation command to apply to the destination file. Requires Puppet version 3.5 or newer. Valid options: a string to be passed to a file resource. Default value: undefined.
--- a/manifests/fragment.pp
+++ b/manifests/fragment.pp
@@ -122,15 +122,16 @@ define concat::fragment(
   }
 
   file { "${fragdir}/fragments/${order}_${safe_name}":
-    ensure  => $safe_ensure,
-    owner   => $fragowner,
-    group   => $fraggroup,
-    mode    => $fragmode,
-    source  => $source,
-    content => $content,
-    backup  => $_backup,
-    replace => true,
-    alias   => "concat_fragment_${name}",
-    notify  => Exec["concat_${target}"]
+    ensure    => $safe_ensure,
+    owner     => $fragowner,
+    group     => $fraggroup,
+    mode      => $fragmode,
+    source    => $source,
+    content   => $content,
+    backup    => $_backup,
+    show_diff => false,
+    replace   => true,
+    alias     => "concat_fragment_${name}",
+    notify    => Exec["concat_${target}"]
   }
 }
--- a/tests/fragment.pp
+++ b/tests/fragment.pp
@@ -1,9 +1,10 @@
 concat { 'testconcat':
-  ensure => present,
-  path   => '/tmp/concat',
-  owner  => 'root',
-  group  => 'root',
-  mode   => '0664',
+  ensure    => present,
+  path      => '/tmp/concat',
+  owner     => 'root',
+  group     => 'root',
+  mode      => '0664',
+  show_diff => true,
 }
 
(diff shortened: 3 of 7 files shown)


HelenCampbell:
Hi,
Thank you for your work with this so far, this seems like a sensible addition to this module. However you would need to have all unit tests passing for us to have a look at this properly.

jdkindy:
@HelenCampbell - I fixed the tests.  Please take a look to see if there is anything else I need to fix.

jdkindy:
Found and removed a whitespace change.

jdkindy:
@HelenCampbell Do I need to submit a new pull request?

Gemfile
```diff
@@ -16,6 +16,7 @@ group :development, :unit_tests do
   gem 'simplecov',               :require => false
   gem 'puppet_facts',            :require => false
   gem 'json',                    :require => false
+  gem 'metadata-json-lint',      :require => false
```
igalic: is this not moduleynced?

jdkindy: I do not know what is meant by 'moduleynced'.  I had to add that line in order to get the tests to pass on my local box.

jdkindy: Removed line locally and retested successfully, so removed line from the commit.  Tests are successful.

bmjen:
I think a `show_diff` param would be quite useful for concat, however I am not a fan of adding the corresponding param to the `concat::fragment`, since the fragment files themselves are supposed to be ingredients to the resulting file vs independently managed file resources. Also, when we do reintroduce concat as a native type, the utility of having a `show_diff` on fragments will become obsolete.

jdkindy:
The reason for adding the show_diff parameter to concat::fragment resources is that diffs of the file resources defined as part of concat::fragment are included in the reports.  Leaving show_diff out of concat::fragment allows secrets to appear in reports.  Puppet runs show diffs of both the resultant concat file resource as well as diffs of the concat::fragment file resources.

Vincent--:
I'd like to have this feature as well (for the same "secret" reason ;-) )

bmjen:
@jdkindy I think I understand the reticence towards show_diff and how it corresponds to the fragment files. However, I am still against the exposure of the show_diff param in the fragment resource, as fragments should be internal resources only. I think an appropriate middle ground would be to remove the show_diff param from the fragment.. and force the fragment file to show_diff=false all the time. 

Thoughts?

Vincent--:
I totally agree ! :-)

jdkindy:
@bmjen I intended for the users to be able to choose fragments if they wished.  However, if the plan is for concat to move to a native type and the concat::fragment type will not have a show_diff, I am ok with setting it to false for all concat::fragment resources.

manifests/fragment.pp
```diff
@@ -26,14 +26,14 @@
...
     $target,
-    $content = undef,
-    $source  = undef,
-    $order   = '10',
-    $ensure  = undef,
-    $mode    = undef,
-    $owner   = undef,
-    $group   = undef,
-    $backup  = undef
+    $content   = undef,
```
bmjen: The alignment here needs to be fixed since `show_diff` was removed.

README.md
```diff
@@ -246,6 +250,10 @@ Specifies whether the fragment should be included in the destination file or dis
 
 Reorders your fragments within the destination file. Fragments that share the same order number are ordered by name. Valid options: a string (recommended) or an integer. Default value: '10'.
 
+#####`show_diff`
```
bmjen: This section needs to be removed.

tests/fragment.pp
```diff
@@ -1,19 +1,20 @@
...
+  group     => 'root',
+  mode      => '0664',
+  show_diff => true,
 }
 
 concat::fragment { '1':
-  target  => 'testconcat',
-  content => '1',
-  order   => '01',
+  target    => 'testconcat',
```
bmjen: alignment.

tests/fragment.pp
```diff
@@ -1,19 +1,20 @@
...
+  content   => '1',
+  order     => '01',
 }
 
 concat::fragment { '2':
-  target  => 'testconcat',
-  content => '2',
-  order   => '02',
+  target    => 'testconcat',
```
bmjen: alignment.

bmjen:
Thanks for the update @jdkindy! A few more comments and we're set for merge!

jdkindy:
I think this commit takes care of all of the issues with the previous one.

bmjen:
Awesome. Thanks for all the work @jdkindy !

Vincent--:
Thks @jdkindy :-)

PR #371, opened 2015-10-28
Tests the fix for pup-1963

bmjen:


spec/acceptance/pup-1963_spec.rb
```diff
@@ -0,0 +1,69 @@
+require 'spec_helper_acceptance'
+
+case fact('osfamily')
+  when 'Windows'
+    command = 'cmd.exe /c echo triggered'
+  else
+    command = 'echo triggered'
```
joshcooper: I think there are beaker helpers for handling this, e.g. `host.echo('triggered')`

PR #374, opened 2015-11-19
Update the line endings for quoted_paths_spec.rb

bwwhite:
This PR converts `quoted_paths_spec.rb` to LF line endings.

In puppetlabs/puppetlabs-concat/pull/373 the `.gitattributes` file was updated so that all .rb files should have LF line endings. `quoted_paths_spec.rb` had CRLF endings, but was not converted and committed with that PR. This means what when cloning master, `quoted_paths_spec.rb` would immediately show as modified, since `.gitattributes` wants to convert it to LF line endings.

I'm admittedly not very knowledgeable about this, I used https://help.github.com/articles/dealing-with-line-endings/ for reference.


bwwhite:
Found this issue because my librarian-puppet install failed with this error:

```
$ librarian-puppet install
Could not checkout https://github.com/puppetlabs/puppetlabs-concat.git at 04356974f72b90a1d0f57346a00e95a717924e43: error: Your local changes to the following files would be overwritten by checkout:
	spec/acceptance/quoted_paths_spec.rb
Please, commit your changes or stash them before you can switch branches.
Aborting
```

bwwhite:
Looks like puppetlabs/puppetlabs-concat/pull/375 was just merged, which does the same thing. This can probably be closed

karyon:
sorry for the confusion caused by my PR's incompleteness...

bmjen:
Thanks for contribution @bwwhite!

PR #386, opened 2016-01-31
Fix the minimum required stdlib version to 4.2.0.

xa4a:
https://github.com/puppetlabs/puppetlabs-concat/commit/f4241b5ffea66e69d9c51fa3ae327fdcd379013f referenced is_bool() from stdlib that was only added in 4.2.0, according to: https://github.com/puppetlabs/puppetlabs-stdlib/blob/04fa5a0cd9255dc2d51af517be7164db9ea0306b/CHANGELOG.md


tphoney:
@bmjen can you have a look at this ? It makes sense the change, but i wonder about backwards compatibility already being broken.

tphoney:
Thanks for the work.

PR #378, opened 2016-01-21
(FM-4046) Update LICENSE file and misc modulesync_config changes

DavidS:
This doesn't change the actually license, but only updates the files to be more in line with ASF's guidelines to applying the license.

PR #388, opened 2016-02-16
(FM-4046) Update to current msync configs [006831f]

DavidS:


NOTICE
```diff
@@ -0,0 +1,20 @@
+concat puppet module
+
+Copyright (C) 2015-2016 Puppet Labs Inc
+Copyright 2012 R.I.Pienaar
```
hunner: @igalic from #378

PR #394, opened 2016-03-12
MODULES-3156: propagate the validate_cmd to the file resource

vicinus:
At the moment `validate_cmd` does nothing because it isn't propagated to the `file` resource.


bmjen:
Thanks @vicinus !

igalic:
eta new release?

PR #397, opened 2016-05-04
(MODULES-3332 ) Correct the path validation.

binford2k:
Prior to this commit, the validation would check for a `Concat_file` with the same **title** as the fragment's **path**. Unfortunately, they're not guaranteed to be the same. The `path` defaults to the `title` but can be passed in by the user.

### DIFF
--- a/lib/puppet/type/concat_fragment.rb
+++ b/lib/puppet/type/concat_fragment.rb
@@ -43,8 +43,8 @@ Puppet::Type.newtype(:concat_fragment) do
   end
 
   autorequire(:file) do
-    unless catalog.resource("Concat_file[#{self[:target]}]")
-      warning "Target Concat_file[#{self[:target]}] not found in the catalog"
+    if catalog.resources.select {|x| x.class == Puppet::Type::Concat_file and x[:path] == self[:target] }.empty?
+      warning "Target Concat_file with path of #{self[:target]} not found in the catalog"
     end
   end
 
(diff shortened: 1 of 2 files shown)


binford2k:
I should also say that, this is an unusual use of `autorequire` since it's not actually requiring anything. It appears that it's being used just for scheduling; to run the check after all resources have been added to the catalog.

asasfu:
I concur with the odd unusual usage, unless that was intention.  I do the same sort of thing but actually utilize the autorequire.  https://github.com/sfu-rcg/puppet-firewalld/blob/master/lib/puppet/type/firewalld_rich_rule.rb#L125-L129

hunner:
Yep, it's specifically because `autorequire` is the only hook we could find to run the check.

hunner:
@binford2k It appears that this change causes an existing test to fail: https://github.com/puppetlabs/puppetlabs-concat/blob/master/spec/acceptance/warnings_spec.rb#L45

hunner:
@binford2k Sorry, could you also update the readme as well? The `target` parameter specifically calls out the concat title.

nibalizer:
+1

hdeheer:
The check looks specifically to fix the case `concat_file {'a_file': path => '/some/path'}` and `concat_fragment {'foo': target => '/some/path'}`
But now we get warnings on the case `concat_file {'a_file': path => '/some/path'}` and `concat_fragment {'foo': target => 'a_file'}`

PR #400, opened 2016-05-18
(MODULES-3332) Correct target validation

hdeheer:
The check should not warn when fragment `target` is based on `concat_file` `name` instead of `path`.
Previous fix only specifically fixed the case:
```
concat_file {'a_file': path => '/some/path'}
concat_fragment {'foo': target => '/some/path'}
```

But it introduced warnings for the case:
```
concat_file {'a_file': path => '/some/path'}
concat_fragment {'foo': target => 'a_file'}
```


DavidS:
Thanks, good catch!

PR #412, opened 2016-08-25
Keep the whole fragment title to sort

sathieu:
... and don't stop at __


sathieu:
Without this, "a__b" and "a__a" may not be sorted correctly.

eputnam:
thanks for the PR! can you please add some tests for this change? something that might fail without this change but would pass with it

PR #421, opened 2016-12-13
[#puppethack] MODULES-3900 add resource info to error messages

aharden:
Also made error messages more consistent and updated relevant spec test.

### DIFF
--- a/spec/unit/defines/concat_fragment_spec.rb
+++ b/spec/unit/defines/concat_fragment_spec.rb
@@ -51,7 +51,7 @@ describe 'concat::fragment', :type => :define do
       let(:params) {{ :target => false }}
 
       it 'should fail' do
-        expect { catalogue }.to raise_error(Puppet::Error, /is not a String/)
+        expect { catalogue }.to raise_error(Puppet::Error, /is not a string/)
       end
     end
   end # target =>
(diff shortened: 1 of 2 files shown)


spec/unit/defines/concat_fragment_spec.rb
```diff
@@ -51,7 +51,7 @@
       let(:params) {{ :target => false }}
 
       it 'should fail' do
-        expect { catalogue }.to raise_error(Puppet::Error, /is not a string/)
+        expect { catalogue }.to raise_error(Puppet::Error, /is not a String/)
```
bmjen: This test validates the `$target` param, which you didn't update.

spec/unit/defines/concat_fragment_spec.rb
```diff
@@ -71,7 +71,7 @@
       let(:params) {{ :content => false, :target => '/etc/motd' }}
 
       it 'should fail' do
-        expect { catalogue }.to raise_error(Puppet::Error, /is not a string/)
+        expect { catalogue }.to raise_error(Puppet::Error, /is not a String/)
```
bmjen: This test validates the `$content` param which was also not changed.

PR #424, opened 2017-01-07
(MODULES-4264) Update for Puppet 4

jeefberkey:
- Updated docs for puppet strings
- Removed deprecated parameters
- Strong type parameters and remove deprecated validate_* functions

### DIFF
--- a/manifests/fragment.pp
+++ b/manifests/fragment.pp
@@ -1,7 +1,6 @@
 # Creates a concat_fragment in the catalogue
 #
 # @param target The file that these fragments belong to
-# @param ensure
 # @param content If present puts the content into the file
 # @param source If content was not specified, use the source
 # @param order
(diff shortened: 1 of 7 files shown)


bmjen:
This is great work @jeefberkey ! I'm going to mark this as milestone 3.0.0.

metadata.json
```diff
@@ -103,11 +103,11 @@
   "requirements": [
     {
       "name": "pe",
-      "version_requirement": ">= 3.7.0 < 2015.4.0"
+      "version_requirement": ">= 2015.4.0 < 2016.4.0"
```
op-ct: Should 2016.4.0 really be the upper bound for this module?

bmjen: Actually.. the whole "pe" block of the requirements is deprecated. We can remove that section and keep just the "puppet" section.

jeefberkey:
Rebased and fixed up the metadata.json, as reviewed.

bmjen:
Thanks @jeefberkey ! We're going to keep this PR around though in the open state, until we are ready to make concat master backwards incompatible.

jeefberkey:
@bmjen I understand! I just don't want it to get too stale.

manifests/fragment.pp
```diff
@@ -1,41 +1,22 @@
...
 #
 define concat::fragment(
-    $target,
-    $ensure  = undef,
-    $content = undef,
-    $source  = undef,
-    $order   = '10',
+  String                  $target,
+  Optional[String]        $content = undef,
+  Variant[String,Integer] $order   = 10,
```
ekohl: Would it make sense to define an allowed regex rather than a String? Then you can drop the ```fail()```in the code.

hunner: Good idea; I tried to do this in #437 but ran into https://tickets.puppetlabs.com/browse/PUP-7440 so the fail couldn't go away.

ekohl:
This needs a rebase now.

jeefberkey:
@hunner This has been superseded or merged elsewhere, correct? If so, I will close this PR.

ekohl:
@jeefberkey your PR goes a bit further in some ways (like removing the deprecated force parameter) but it's been mostly superseded yes.

hunner:
@jeefberkey Yes thanks for all of your work! https://github.com/puppetlabs/puppetlabs-concat/pull/437 is literally your 3 commits (before you added the 4th one here) plus one from me to add back deprecation stuff.

PR #428, opened 2017-01-27
tdevelioglu/refurbish

tdevelioglu:
This is a combined PR for the following issues:

[MODULES-4349 - Concat: Fix obsolete Rubocop config](https://tickets.puppetlabs.com/browse/MODULES-4349)
[MODULES-4350 - Concat: Fix path parameter not being a namevar.](https://tickets.puppetlabs.com/browse/MODULES-4350)
[MODULES-4351 - Concat: fix :order parameter accepting arbitrary values in concat_file](https://tickets.puppetlabs.com/browse/MODULES-4351)
[MODULES-4352 - Concat: Restrict parameter :replace to values accepted by file](https://tickets.puppetlabs.com/browse/MODULES-4352)
[MODULES-4359 - Concat: Fix native types so they can be used directly.](https://tickets.puppetlabs.com/browse/MODULES-4359)

### DIFF
--- a/lib/puppet/type/concat_file.rb
+++ b/lib/puppet/type/concat_file.rb
@@ -11,7 +11,7 @@ Puppet::Type.newtype(:concat_file) do
       Concat_fragment <<| tag == 'unique_tag' |>>
 
       concat_file { '/tmp/file':
-        tag            => 'unique_tag', # Mandatory
+        tag            => 'unique_tag', # Optional. Default to undef
         path           => '/tmp/file',  # Optional. If given it overrides the resource name
         owner          => 'root',       # Optional. Default to undef
         group          => 'root',       # Optional. Default to undef
@@ -67,7 +67,12 @@ Puppet::Type.newtype(:concat_file) do
 
   newparam(:backup) do
     desc "Controls the filebucketing behavior of the final file and see File type reference for its use."
-    defaultto 'puppet'
+
+    validate do |value|
+      unless [TrueClass, FalseClass, String].include?(value.class)
+        raise ArgumentError, 'Backup must be a Boolean or String'
+      end
+    end
   end
 
   newparam(:replace, :boolean => true, :parent => Puppet::Parameter::Boolean) do
@@ -77,6 +82,12 @@ Puppet::Type.newtype(:concat_file) do
 
   newparam(:validate_cmd) do
     desc "Validates file."
+
+    validate do |value|
+      unless value.is_a?(String)
+        raise ArgumentError, 'Validate_cmd must be a String'
+      end
+    end
   end
 
   newparam(:ensure_newline, :boolean => true, :parent => Puppet::Parameter::Boolean) do
@@ -85,28 +96,51 @@ Puppet::Type.newtype(:concat_file) do
   end
 
   # Inherit File parameters
-  newparam(:selinux_ignore_defaults) do
-  end
+  newparam(:selinux_ignore_defaults, :boolean => true, :parent => Puppet::Parameter::Boolean)
 
   newparam(:selrange) do
+    validate do |value|
+      raise ArgumentError, 'Selrange must be a String' unless value.is_a?(String)
+    end
   end
 
   newparam(:selrole) do
+    validate do |value|
+      raise ArgumentError, 'Selrole must be a String' unless value.is_a?(String)
+    end
   end
 
   newparam(:seltype) do
+    validate do |value|
+      raise ArgumentError, 'Seltype must be a String' unless value.is_a?(String)
+    end
   end
 
   newparam(:seluser) do
+    validate do |value|
+      raise ArgumentError, 'Seluser must be a String' unless value.is_a?(String)
+    end
   end
 
-  newparam(:show_diff) do
-  end
+  newparam(:show_diff, :boolean => true, :parent => Puppet::Parameter::Boolean)
   # End file parameters
 
   # Autorequire the file we are generating below
+  # Why is this necessary ?
   autorequire(:file) do
     [self[:path]]
+  end
+
+  def fragments
+    # Collect fragments that target this resource by path, title or tag.
+    @fragments ||= catalog.resources.map do |resource|
+      next unless resource.is_a?(Puppet::Type.type(:concat_fragment))
+
+      if resource[:target] == self[:path] || resource[:target] == title ||
+        (resource[:tag] && resource[:tag] == self[:tag])
+        resource
+      end
+    end.compact
   end
 
   def should_content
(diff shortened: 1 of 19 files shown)


tdevelioglu:
I know this is a rather large commit, would you like me to split it up into a number of smaller PR's instead ?

lib/puppet/type/concat_file.rb
```diff
@@ -58,68 +54,98 @@ def exists?
   end
 
   newparam(:order) do
-    desc "Controls the ordering of fragments. Can be set to alphabetical or numeric."
-    defaultto 'numeric'
+    desc "Controls the ordering of fragments. Can be set to :alpha or :numeric."
```
hunner: Docs should not have :'s for the argument values.

lib/puppet/type/concat_file.rb
```diff
@@ -58,68 +54,98 @@ def exists?
...
+
+    validate do |value|
+      unless [TrueClass, FalseClass, String].include?(value.class)
+        raise ArgumentError, 'Backup must be a Boolean or String'
+      end
+    end
   end
 
-  newparam(:replace) do
+  newparam(:replace, :boolean => true, :parent => Puppet::Parameter::Boolean) do
```
hunner: I think if you make this `:boolean => true` then you don't need the parent?

lib/puppet/type/concat_file.rb
```diff
@@ -58,68 +54,98 @@ def exists?
...
+    desc "Controls the ordering of fragments. Can be set to :alpha or :numeric."
+
+    newvalues(:alpha, :numeric)
+
+    defaultto :numeric
   end
 
   newparam(:backup) do
     desc "Controls the filebucketing behavior of the final file and see File type reference for its use."
-    defaultto 'puppet'
```
hunner: Why the removal of this default? Is that backwards incompatible?

lib/puppet/type/concat_file.rb
hunner: Fixed in #433

lib/puppet/type/concat_file.rb
hunner: Oh, I see that the defaultto 'puppet' is set in lib/puppet/type/file.rb in puppet's source.

lib/puppet/type/concat_file.rb
hunner: Fixed in #433

lib/puppet/type/concat_file.rb
```diff
@@ -85,40 +92,56 @@ def exists?
   end
 
   # Inherit File parameters
-  newparam(:selinux_ignore_defaults) do
-  end
+  newparam(:selinux_ignore_defaults, :boolean => true, :parent => Puppet::Parameter::Boolean)
```
hunner: No parent needed.

hunner:
#433 brings in the first four commits (yay! Thank you!) with a few tweaks, so the review is just for the last commit about tag/path.

lib/puppet/type/concat_file.rb
```diff
@@ -58,68 +54,98 @@ def exists?
...
+    end
   end
 
-  newparam(:show_diff) do
-  end
+  newparam(:show_diff, :boolean => true, :parent => Puppet::Parameter::Boolean)
   # End file parameters
 
   # Autorequire the file we are generating below
+  # Why is this necessary ?
```
hunner: ¯\_(ツ)_/¯ I don't recall. It seems like the generated file will be contained by the concat_file resource and wouldn't need this.

lib/puppet/type/concat_file.rb
hunner: Well what do you know, *apparently* it does need a parent. Is this because it's a parameter, and only properties don't need a parent?

lib/puppet/type/concat_file.rb
tdevelioglu: :boolean => true just adds a `param?` method to the type, you still need the parent for value checking and coercion.

lib/puppet/type/concat_file.rb
tdevelioglu: Technically no, they're not any different. It's just that setting Puppet::Property::Boolean as a parent of a property breaks the property which is why often you see the same bits copied directly into the property.

lib/puppet/type/concat_file.rb
tdevelioglu: I see. The intent of my change was to allow the types to be used more easily directly but the tag normalization was making it tricky. By removing the tag and using the path or title as the target my hope was to get around it.

It *is* very valuable to be able to use the types without the wrappers however. My testing shows a ~10% reduction in total resource count, and ~20% decrease (2.5M -> 2M) in catalog size on one of our concat heavy roles. This adds up real quickly in real-world numbers when you consider an environment of 20-30k hosts which make heavily use of concat, especially on the inventory/puppetdb side of things, since all file content effectively gets duplicated when it gets passed through the wrappers.

Perhaps I can re-factor it to collect on both tag *and* target, title, how does that sound ?

tdevelioglu: I took another look at this and `concat_fragment` already *has* to know what the target will be, it can not be declared otherwise, see:

https://github.com/puppetlabs/puppetlabs-concat/blob/master/lib/puppet/type/concat_fragment.rb#L45-L49

tdevelioglu: Ah, I missed that it only emits a warning but it makes it seem as if the intent is that fragments should target concat_file by path or title.

tdevelioglu: It does raise the question what the actual purpose of the target parameter is then, because it's not used for anything except emitting that warning.

lib/puppet/type/concat_file.rb
hunner: I removed the parent, then tests failed. It apparently accepted lots of truthy values because of `:boolean => true`, but didn't sanitize them on the way back. I didn't track down as to why yet though. It could have just been poorly defined behavior checking in the unit tests.

Oh how I wish we had actual manifest/catalog type & provider unit testing. Someday.

lib/puppet/type/concat_file.rb
hunner: I think the target exists for backwards-compatibility reasons.

tdevelioglu:
@hunner I've refactored the types to work with both tag and target only. Can you take a look and let me know what you think?

spec/acceptance/native/backup_spec.rb
```diff
@@ -0,0 +1,115 @@
+require 'spec_helper_acceptance'
```
hunner: It looks like all of the acceptance tests except `warn_header_spec.rb` were copied from spec/acceptance/ to spec/acceptance/native/ but the originals were not modified. Was this on purpose?

tdevelioglu: The examples in spec/acceptance/native/ test the native types directly. They're modified originals.

PR #429, opened 2017-02-21
Invalid tag fix

fuero:
concat doesn't like the @ symbol in file paths.

This fixes this and amends the unit test to check for this.


tphoney:
Thanks for the great fix @fuero, simple and elegant.

PR #430, opened 2017-03-01
(MODULES-4474) Drop autorequire of fragments in concat_file

tdevelioglu:
The concat_file type defines an autorequire of related fragments which
can unnecessarily lead to dependency cycles.

It's not evident why fragments exactly need to be required as the type
seems to work fine without it.

This commit drops the autorequire that collects fragments in
concat_file to prevent these dependency cycles from happening.


bmjen:
Thanks @tdevelioglu ! I believe this was a remnant of the T&P changes before we moved some of the magic to eval_generate.

PR #437, opened 2017-03-27
(MODULES-4264) Update for puppet 4 data types

hunner:
The data types for validation are using the puppet 4 types with
validate_legacy to check for any legacy types that will be removed on
the next release. This should be a backwards-compatible change except
that it uses puppet 4's data types.

### DIFF
--- a/manifests/fragment.pp
+++ b/manifests/fragment.pp
@@ -1,7 +1,6 @@
 # Creates a concat_fragment in the catalogue
 #
 # @param target The file that these fragments belong to
-# @param ensure
 # @param content If present puts the content into the file
 # @param source If content was not specified, use the source
 # @param order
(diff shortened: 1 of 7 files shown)


hunner:
Actually I should get this on top of #424. Oops.

manifests/fragment.pp
```diff
@@ -15,27 +15,19 @@
...
 define concat::fragment(
-    $target,
-    $ensure  = undef,
-    $content = undef,
-    $source  = undef,
-    $order   = '10',
+    Variant[String, Stdlib::Compat::String]                                   $target,
+                                                                              $ensure  = undef,
+    Optional[Variant[String, Stdlib::Compat::String]]                         $content = undef,
+    Optional[Variant[String, Stdlib::Compat::String, Array]]                  $source  = undef,
```
HelenCampbell: Add Stdlib::Compat::Array to the variant also?

manifests/fragment.pp
```diff
@@ -15,27 +15,19 @@
...
+    Variant[String, Stdlib::Compat::String, Integer, Stdlib::Compat::Integer] $order   = '10',
 ) {
-  validate_string($target)
```
HelenCampbell: Instead of straight up removing these, use the validate_legacy function here: https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/puppet/functions/validate_legacy.rb
A description of why/usage: https://github.com/puppetlabs/puppetlabs-stdlib#validate_legacy

manifests/fragment.pp
hunner: Okay. And you too on https://github.com/puppetlabs/puppetlabs-apt/pull/659/files#diff-f0fb8c712e70efd75993654e78e8bd59R8 then :D

metadata.json
```diff
@@ -103,7 +103,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 3.0.0 < 5.0.0"
+      "version_requirement": ">= 4.5.2 < 5.0.0"
```
bastelfreak: shouldn't this be 4.6.1? As far as I know this is the first puppet4 version that is considered stable. Vox Pupuli bumped everything to 4.6.1. All based on [this statement](https://github.com/voxpupuli/community-triage/blob/master/modules/notes/2017-01-05.md#discussion)

hunner: Preferably 4.6.1 because yes it works better (mostly around create_resources bugs like https://tickets.puppetlabs.com/browse/PUP-6230), but the 4.5 series was the first that fixed data-in-modules (https://tickets.puppetlabs.com/browse/PUP-6230). I'll check with the gang.

hunner: PE 2016.4 (LTS) includes 4.7 but the previous release 2016.2 (they skipped .3) includes 4.5.2... but according to https://docs.puppet.com/pe/latest/overview_getting_support.html 2016.2 is EOL April 30, 2017 so we could probably just jump straight to 4.7.0 support-wise, and use the older puppet-3 versions of modules for supporting 2016.2 and earlier until they EOL.

tphoney:
👍

PR #443, opened 2017-04-11
[concat] Puppet::Parameter::Boolean yields error on puppet 3.2.

truthbk:
So while running my spec tests on my module, I encountered an issue when testing for puppet 3.2 (all other supported versions seemed to be fine).

The error reported was: 
```
Puppet::Error:                                                                                                                                                                  
       Could not autoload puppet/type/concat_file: uninitialized constant Puppet::Parameter::Boolean on (redacted)
```

Since puppet >=3.0 is supported, I believe this is unexpected behavior and should somehow be fixed. I was able to get around the problem by applying the same fix the guys in the CPAN module used (https://github.com/meltwater/puppet-cpan/commit/59a1acfe5e560687da923dbf345bad829d407d35).

This PR is likely not complete as it doesn't address the spec tests define in spec/unit/type/concat_file_spec.rb, as I'm not sure what that entails. I would've actually preferred to open an issue instead of a PR, but you guys don't have issues enabled :)

Anyways, don't hesitate to let me know if you need anything else from me. 

Thanks!

### DIFF
--- a/lib/puppet/type/concat_file.rb
+++ b/lib/puppet/type/concat_file.rb
@@ -70,7 +70,7 @@ Puppet::Type.newtype(:concat_file) do
     defaultto 'puppet'
   end
 
-  newparam(:replace, :boolean => true, :parent => Puppet::Parameter::Boolean) do
+  newparam(:replace, :boolean => true) do
     desc "Whether to replace a file that already exists on the local system."
     defaultto :true
   end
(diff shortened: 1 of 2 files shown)


truthbk:
Another relevant commit, this one in puppetlabs-stdlib: https://github.com/puppetlabs/puppetlabs-stdlib/commit/a7adcda803abe82e6a16e2410c10d58abedbd82d

ekohl:
Master has dropped support for puppet 3 so I think this is unlikely to be accepted.

PR #447, opened 2017-04-14
(FM-6112) release prep for 3.0.0 and 4.0.0

eputnam:


CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## Supported Release 3.0.0
+### Summary
+
+This is a backwards incompatible major release.
```
hunner: We should probably describe what is incompatible.
> This major release drops puppet 3 compatibility, but is otherwise backwards compatible.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
...
+
+### Features
+
+- Docs updated for puppet-strings
+- Adds Puppet 4 data types and deprecation warnings
```
hunner: This needs a documented section for the breaking change in addition to the list of added features.

IMHO according to http://keepachangelog.com/ we should have `### Added` for new features, and `### Changed` for the backwards-incompatible changes.

(And `### Fixed` for bugfixes, if there were any.)

eputnam: historically we've only had `### Features` and `### Bugfixes` sections. I'm cool to change but we should make sure it happens across all modules. 

Also, I'm looking at the diff from 2.2.1 which is pretty slim. Should I be diff-ing from the release before that?

PR #456, opened 2017-06-23
(MODULES-5122) Verify Puppet >= v3.5 before adding 'validate_cmd' param

guilledipa:
This commit makes fixes a catalog compilation failure ("Invalid parameter validate_cmd") in systems running Puppet versions <3.5 using concat 1.2.5.
init.pp does not check if the puppet version is >=3.5 and adds 'validate_cmd' parameter which breaks the catalog compilation. 'validate_cmd' parameter should only be added if puppet version >= 3.5.

Also added more meaningful comment to the if statement where the parameter is added.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -194,8 +194,8 @@ define concat(
       backup  => $backup,
     }
 
-    # Only newer versions of puppet 3.x support the validate_cmd parameter
-    if $validate_cmd {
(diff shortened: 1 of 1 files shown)


tphoney:
Thanks for the PR, however i cannot merge it. With the release of Concat 4.0 only puppet 4 and above is supported. The additional logic you have added is unnecessary.

Thanks again, and apologies for the inconvenience.

PR #458, opened 2017-06-26
is_string is deprecated. Please use puppet language

tritrimax:
Warning: This method is deprecated, please use match expressions with Stdlib::Compat::String instead. They are described at https://docs.puppet.com/puppet/latest/reference/lang_data_type.html#match-expressions. at ["concat/manifests/fragment.pp", 18]:
==> myserver.local:    (stdlib/lib/puppet/functions/deprecation.rb:25:in `deprecation')

### DIFF
--- a/manifests/fragment.pp
+++ b/manifests/fragment.pp
@@ -15,7 +15,7 @@ define concat::fragment(
 ) {
   $resource = 'Concat::Fragment'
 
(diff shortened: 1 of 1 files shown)


tphoney:
@HelenCampbell can you have a look at this please 🥇

HelenCampbell:
@tritrimax Thanks!

PR #461, opened 2017-07-03
Don't call join on a potential string when failing

antaflos:
When `fail()`ing in case a fragment source is not found don't call `join` directly on `r[:source]` because it might very well be a string, not an array. Wrap `r[:source]` in a (compacted, flattened) array before calling `join`.

This avoids confusing error messages like `Failed to generate additional resources using 'eval_generate': undefined method 'join' for "foo":String` when a fragment source cannot be found. See also https://tickets.puppetlabs.com/browse/MODULES-4140.

### DIFF
--- a/lib/puppet/type/concat_file.rb
+++ b/lib/puppet/type/concat_file.rb
@@ -183,7 +183,7 @@ Puppet::Type.newtype(:concat_file) do
           break
         end
       end
-      self.fail "Could not retrieve source(s) #{r[:source].join(", ")}" unless @source
+      self.fail "Could not retrieve source(s) #{[r[:source]].flatten.compact.join(", ")}" unless @source
       tmp = Puppet::FileServing::Content.indirection.find(@source)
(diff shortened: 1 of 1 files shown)


tphoney:
@antaflos thanks for the PR this seems like an excellent addition, could you possibly add some tests for this.

antaflos:
@tphoney I could, but according to https://tickets.puppetlabs.com/browse/MODULES-4140?focusedCommentId=459578&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-459578 this PR is probably moot anyway and a better fix is setting `:array_matching => :all` for `concat_fragment`.

tphoney:
Should we close this Pr then ? thanks for taking the time to look at this.

tphoney:
Hi @antaflos , thanks for the work you have put into this pull request. However due to inactivity and the reasons your outlined this pull request is being closed.

PR #465, opened 2017-08-21
Move tests to follow the rpsec-puppet pattern

ekohl:
When defines are in spec/defines then rspec-puppet automatically includes the define checks according to https://github.com/rodjek/rspec-puppet/blob/3c16670252f3cf309037f1e0cd3fa71f903bef68/lib/rspec-puppet/example.rb#L22


tphoney:
Many thanks @ekohl

PR #467, opened 2017-10-09
Added Debian Stretch (9) to the supported list

hp197:
Should there be a discussion first if you want to support Debain Stretch?
In any way, Stretch is the new stable for Debian and I dont see any reason not to support it.

We run and tested it on Debian Stretch and we haven't found any issue as of yet.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -57,7 +57,8 @@
       "operatingsystemrelease": [
         "6",
         "7",
-        "8"
+        "8",
+        "9"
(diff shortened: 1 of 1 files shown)


willmeek:
Thank you for your contribution!

This was rolled up and merged along with removal of Debian 6 with pull request https://github.com/puppetlabs/puppetlabs-concat/pull/468

As such this Pull Request is now closed.

PR #469, opened 2017-10-12
Format force tests

willmeek:


tphoney:
Great work @WhatsARanjit  !!

johanfleury:
Hi.

For a reason I can't find, this changes make Puppet fail with this message: `Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: no parameter named 'format' at [...]/modules/concat/manifests/init.pp:94 on Concat_file[/etc/ssh/sshd_config] at [...]/modules/concat/manifests/init.pp:94 on node foo`.

The call to concat can be found here: <https://github.com/saz/puppet-ssh/blob/master/manifests/server/config.pp#L16>.

Any idea of what could cause this error?

WhatsARanjit:
@johanfleury No worries.  The `concat` defined type calls the `concat_file` native type.  Likely the new manifest was deployed but an older version of the type was cached.  You can force an update like this:

https://puppet.com/docs/puppet/5.2/environment_isolation.html#generate-types

willmeek:
Thanks for reporting @johanfleury and @WhatsARanjit 

This is a common problem when updating a module that changes parameters for a native provider (see the [upstream Puppet Labs](https://tickets.puppetlabs.com/browse/SERVER-94) issue). Please try restarting the puppet master (and potentially any affected puppet agents) and try re-running puppet again.

Otherwise to reproduce this issue I would ask that you provide detail on the following Jira ticket: https://tickets.puppetlabs.com/browse/MODULES-5802

In particular:

- Module and version ( as reported by puppet module list )
- Manifests used
- Puppet version
- Ruby Version
- Any applicable logs

When we have enough information to replicate this scenario, we can look to fix if necessary.

Thank you.

PR #485, opened 2018-01-22
Refactor fragment sorting

paran1:


hunner:
Wow nice! This is much better.

HelenCampbell:
We are also seeing this and have had a ticket raised showing the same errors as those shown by @EmilienM , which you can see here: https://tickets.puppetlabs.com/browse/MODULES-6545
I'm going to try and look into this issue and see if I can find out what's causing it. Concat is due a release and we can't until this is addressed.

PR #488, opened 2018-02-21
Revert "Refactor fragment sorting"

pmcmaw:
`Reverts` puppetlabs/puppetlabs-concat#485


pmcmaw:
Hey @paran1,

In order to provide a quick fix for concat it has been decided the best method is to revert your PR #485  as we are currently working on releasing this module and we do not want to make a release that will break set ups. We really appreciate the time you have taken to actually put in this PR and are sorry that we have decided the best way to move forward with our release is to revert this.

Again, apologies.
Paula

PR #493, opened 2018-03-22
Handle concat_file source when not an array

vicinus:
Ensure that the source parameter is an array in the error message so that join
works.


vicinus:
As it's only a one liner and a very minor fix i didn't create a ticket at puppetlabs. I hope this is ok.

PR #494, opened 2018-03-22
(MODULES-6905) Propagate all intended metaparameters to file resource

seanmil:


hunner:
Thanks! This is pretty close to #492 which is already merged and getting ready to go out in 4.2.1. Looks like you took more care in your testing though.

PR #517, opened 2018-08-24
(MODULES-7717) ensure_newline uses unix line ending on windows

tkishel:
Prior to this commit, ensure_newline used '\n' everywhere.
With this commit, ensure_newline uses '\r\n' on Windows.


tkishel:
I just realized that the name of the commit reads differently depending upon how you pronounce the underscore in ensure_newline :|

PR #519, opened 2018-09-09
(MODULES-5124) Add support for JSON arrays

johanfleury:
Hi.

I'm working on a module and I need to concat multiple hashes into a JSON array.

I tried to do my best to implement this in concat, but I'm not really a Ruby guy, so I'm not sure this is the correct way to do so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,7 +406,7 @@ Default value: `undef`.
 
 Data type: String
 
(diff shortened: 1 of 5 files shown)


david22swan:
screen shot 2018-11-05 at 2 36 47 pm

PR #521, opened 2018-09-10
(MODULES-7790) Fix modifying frozen string literal

MikaelSmith:
In Puppet 6, PUP-7141 changed the 3x runtime to stop duplicating
strings. Providers will now usually receive frozen strings. Update the
concat_file provider to create a new copy of a string if it needs to add
to it rather than attempt to mutate the original.


MikaelSmith:
This is less ideal under Puppet 5, where that string isn't frozen. We could test for it first if that's a concern.

PR #530, opened 2018-10-15
Revert "(MODULES-7717) ensure_newline uses unix line ending on windows"

ekinanp:
This reverts commit 0a6e416e37b5493825b22e6e803deffbda26633d to unblock
the agent suite pipelines until MODULES-8088 is fixed.

### DIFF
--- a/lib/puppet/type/concat_file.rb
+++ b/lib/puppet/type/concat_file.rb
@@ -255,8 +255,7 @@ Puppet::Type.newtype(:concat_file) do
     end
 
     if self[:ensure_newline]
-      newline = Puppet::Util::Platform.windows? ? "\r\n" : "\n"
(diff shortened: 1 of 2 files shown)


glennsarti:
Agreed. This is not the fault of the concat module

pmcmaw:
Something to note here is that the code in relation to this revert PR doesn't actually run without failures on Windows. This initial change has broken our pipelines. (eg https://github.com/puppetlabs/puppetlabs-concat/commit/0a6e416e37b5493825b22e6e803deffbda26633d fails on Windows in our pipelines).

Adding @david22swan for visibility as he may be able to provide more information on this or will be working on a fix.

ekinanp:
Only put this up here to get the agent CI passing. If it can be fixed soon, then this revert is not needed.

DavidS:
#531 is addressing this problem without losing functionality.

PR #536, opened 2018-11-29
(FM-7341) - Added REFERENCE.md and updated documentation

eimlav:


lib/puppet/type/concat_fragment.rb
```diff
@@ -1,48 +1,62 @@
...
+      Manages the fragment.
+
+    @example
+      # The example is based on exported resources.
+
+      concat_fragment { \"uniqe_name_${::fqdn}\":
+        tag => 'unique_name',
+        order => 10, # Optional. Default to 10
+        content => 'some content' # OR
+        # content => template('template.erb')
```
tphoney: is this meant to be a comment ?

eimlav: Yea the original example had both content parameters defined to show the alternative options you could choose. However since that is syntactically incorrect and would have thrown an error I just commented one of them. I can change it if it seems confusing.

PR #538, opened 2018-11-30
{WIP} Solid waffle DO NOT MERGE

tphoney:


.fixtures.yml
```diff
@@ -1,6 +1,9 @@
 fixtures:
   repositories:
-    'stdlib': 'git://github.com/puppetlabs/puppetlabs-stdlib.git'
-    'translate': 'git://github.com/puppetlabs/puppetlabs-translate.git'
+    facts: 'git://github.com/puppetlabs/puppetlabs-facts.git'
```
MikaelSmith: I take it facts/puppet_agent are required by Bolt within Solid Waffle?

tphoney: that is correct

PR #539, opened 2018-12-03
(FM-7350) - Minor OS cleanup

eimlav:


eimlav:
Closing in favour of maintenance work carried out in #538

PR #544, opened 2018-12-12
(maint) - Doc Fix

david22swan:


clairecadman:
@david22swan Where do these links go? Are you adding these sections in later?

david22swan:
@clairecadman Ah, meant to close this. Created after I saw that the pr of yours that I merged removed them, but forgot to double check if it was actually an error.

PR #569, opened 2019-02-20
Update README.md to match puppet style guide

mookie-:
I think it's easier to read like this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ class motd {
     mode  => '0644'
   }
 
-  concat::fragment{ 'motd_header':
+  concat::fragment { 'motd_header':
     target  => $motd,
     content => "\nPuppet modules on this server:\n\n",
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -60,30 +60,33 @@ class motd {
...
     target => $motd,
     source => '/etc/motd.local',
     order  => '15'
   }
 }
 
 # let other modules register themselves in the motd
-define motd::register($content="", $order='10') {
+define motd::register (
+  $content="",
```
natemccurdy: There should be spaces around the `=` for all of these parameters as well:
```
define motd::register (
  $content = '',
  $order   = '10',
) {
```

mookie-: @natemccurdy Thank you for the hint. I've updated the PR.

eimlav:
@clairecadman does this look good to you?

PR #584, opened 2019-05-22
FM-8049 - add redhat8 support

lionce:


lionce:
green build: 
https://pipelines.puppet.com/team-modules/builds/550857

eimlav:
@lionce There appears to have been an error on Pipelines win-10-pro-x86_64 Any idea why this wouldn't flag the build as failure?

PR #623, opened 2020-01-03
Fix newline handling at the end of fragments

SimonPe:
`$` matches both end of string and end of line.
so `/\n$/` matches `\n\n` anywhere in the string.
while `/\n\Z/` only matches `\n` at the end of the string.

```
irb(main):021:0> "a\n\nb" =~ /\n$/
=> 1
irb(main):022:0> "a\n\nb" =~ /\n\Z/
=> nil
irb(main):023:0> "a\n\nb\n" =~ /\n$/
=> 1
irb(main):024:0> "a\n\nb\n" =~ /\n\Z/
=> 4
```


SimonPe:
the one failed travis job failed on `bundle exec rake 'litmus:install_agent[puppet6]'` which I don't think this code is involved in.

PR #627, opened 2020-01-29
Add support for Deferred function in concat fragment

baurmatt:
This allows the `content` parameter of `concat::fragment` to be an `Deferred` function.


daianamezdrea:
Hello @baurmatt , 
This looks good , thank you for your contribution !

PR #643, opened 2020-05-18
pdksync - Use abs instead of vmpooler to provision test resources

carabasdaniel:
Use abs instead of vmpooler to provision test resources
pdk version: `1.17.0`


DavidS:
Does this, too, need pinning of bolt ala IAC-811?

sheenaajay:
Resolved conflicts and rebased with master.

PR #660, opened 2020-09-14
Resolve puppet-lint warnings

jcpunk:
I like to run puppet-lint against my whole tree every so often.  It flagged these items.


daianamezdrea:
Hi @jcpunk, this looks good! Thank you for your contribution!

PR #678, opened 2020-12-16
Update to puppet-module-gems 1.0, pdk-templates and new rubocop

DavidS:


DavidS:
rebased and re-running tests with new litmus 0.22.0

PR #750, opened 2023-01-16
PDK update test repos

GSPatton:


chelnak (approved):

Gemfile
```diff
@@ -14,8 +14,12 @@ def location_for(place_or_version, fake_version = nil)
 end
 
 group :development do
-  gem "json", '~> 2.0',                                require: false
-  gem "voxpupuli-puppet-lint-plugins", '~> 3.0',       require: false
+  gem "json", '= 2.1.0',                               require: false if Gem::Requirement.create(['>= 2.5.0', '< 2.7.0']).satisfied_by?(Gem::Version.new(RUBY_VERSION.dup))
+  gem "json", '= 2.3.0',                               require: false if Gem::Requirement.create(['>= 2.7.0', '< 3.0.0']).satisfied_by?(Gem::Version.new(RUBY_VERSION.dup))
+  gem "json", '= 2.5.1',                               require: false if Gem::Requirement.create(['>= 3.0.0', '< 3.0.5']).satisfied_by?(Gem::Version.new(RUBY_VERSION.dup))
+  gem "json", '= 2.6.1',                               require: false if Gem::Requirement.create(['>= 3.1.0', '< 3.1.3']).satisfied_by?(Gem::Version.new(RUBY_VERSION.dup))
+  gem "json", '= 2.6.3',                               require: false if Gem::Requirement.create(['>= 3.2.0', '< 4.0.0']).satisfied_by?(Gem::Version.new(RUBY_VERSION.dup))
```
ekohl, line 21: This really is hard to read. Can't this be a tree of if/else?
```suggestion
if Gem::Requirement.create(['>= 2.5.0', '< 2.7.0']).satisfied_by?(Gem::Version.new(RUBY_VERSION.dup))
  gem "json", '= 2.1.0',                               require: false 
elsif  Gem::Requirement.create(['>= 2.7.0', '< 3.0.0']).satisfied_by?(Gem::Version.new(RUBY_VERSION.dup))
  gem "json", '= 2.3.0',                               require: false
elsif Gem::Requirement.create(['>= 3.0.0', '< 3.0.5']).satisfied_by?(Gem::Version.new(RUBY_VERSION.dup))
  gem "json", '= 2.5.1',                               require: false
if Gem::Requirement.create(['>= 3.1.0', '< 3.1.3']).satisfied_by?(Gem::Version.new(RUBY_VERSION.dup))
  gem "json", '= 2.6.1',                               require: false
elsif Gem::Requirement.create(['>= 3.2.0', '< 4.0.0']).satisfied_by?(Gem::Version.new(RUBY_VERSION.dup))
  gem "json", '= 2.6.3',                               require: false
end
```

chelnak: I'm not entirely sure how this would work at the moment. These changes were generated from the config supplied by pdk-templates.

There may be a way to make it look nicer though.. am open to any suggestions before we push this out to all of our modules.

ekohl: One option I'd consider is maybe way easier: drop the whole dependency on `json`. We don't have it in our Vox Pupuli modules and the `json` gem is a default gem (since Ruby 1.9 I believe). So my suggestion would be to drop it and then see if it still passes.

chelnak: So this actually a PDK thing.

Each pin is version of JSON that comes with the respective Ruby.

Their presence in module Gemfiles stops PDK trying to build the required native extensions when you initialise a new module or update.

It's something I want to see if I can fix in a future version because it's less than ideal.

ekohl: Looking at a `Gemfile.lock` from https://github.com/voxpupuli/puppet-example/blob/master/Gemfile it looks like `simplecov` and `coveralls` pull in `json` as a dependency. Personally I think they're mostly useless and if I drop both `coveralls` and `simplecov` I don't see it pull in json at all anymore.

That's the route I'd suggest: figure out why `json` is pulled in in the first place and get rid of it.

If we apply that to `puppetlabs-concat`, we can remove it and `bundle install` doesn't install `json` anymore.

chelnak: Yeah I noticed that too.. my assumption was that the issue I mention [above](https://github.com/puppetlabs/puppetlabs-concat/pull/750#discussion_r1072531117) is a side effect of the way everything is vendored in to the PDK packages.

I'm more than happy to explore this though because if it removes the need for those pins then i'd say that is a win.

Sadly though, these changes are going to appear across our modules to complement the release of PDK 2.6.0.. but we can easily rip them out again if we find a fix!

Thanks for the suggestion !

PR #757, opened 2023-03-10
Allow content parameter of concat_fragment to be Sensitive

### LINKED ISSUE #742: Support for content with sensitive type
## Use Case
If there is sensitive content included in a file (like a password), I don’t want it to be displayed in the log.

## Describe the Solution You Would Like
Support for the `Sensitive` type in the content:
```
class secret_within_content {
  $secret = Sensitive('mysecret')
  concat { '/tmp/example.conf':  }
  concat::fragment { 'secret' :
    target  => '/tmp/example.conf',
    content => $secret,
  }
}

class with_inline_epp {
  $secret = Sensitive('mysecret')
  $content = inline_epp("secret = <%= $s %>", { s => $secret })
  concat { '/tmp/example.conf':  }
  concat::fragment { 'secret' :
    target  => '/tmp/example.conf',
    content => $content,
  }
}
```

## Describe Alternatives You've Considered
The alternative for now it to use `show_diff => false`, which don’t show any diff at all, but I would like to only remove the data that is sensitive, not the entire diff from the logs

## Additional Context
I’ve tried adding `Sensitive` in `fragment.pp` on line 47:
```
46  if versioncmp($clientversion, '6.0') >= 0 and versioncmp($_serverversion, '6.0') >= 0 {
47    assert_type(Optional[Variant[String, Deferred, Sensitive]], $content)
48  } else {
49    assert_type(Optional[String], $content)
50  }
```
But then there is a warning and the diff still show the sensitive content: `Warning: /Concat_fragment[secret]: Unable to mark 'content' as sensitive: content is a parameter and not a property, and cannot be automatically redacted.`  I would like to offer a pull request on this, but i’m unsure where I would go from here...  Is a complete provider has to be written for this to work?

baurmatt:
Fixes #742.


manifests/fragment.pp
```diff
@@ -40,9 +40,9 @@
     default => $serverversion,
   }
   if versioncmp($clientversion, '6.0') >= 0 and versioncmp($_serverversion, '6.0') >= 0 {
-    assert_type(Optional[Variant[String, Deferred]], $content)
+    assert_type(Optional[Variant[Sensitive[String], String, Deferred]], $content)
   } else {
-    assert_type(Optional[String], $content)
+    assert_type(Optional[Sensitive[String], String], $content)
```
ekohl, line 45: Did Puppet 5 have a `Sensitive` type? Though I also wonder if by now it couldn't be dropped.

ekohl (commented):
Recently I've been thinking about a type alias `Stdlib::File::Content` that's an alias for what the `file` type accepts for the `content` parameter so I opened https://github.com/puppetlabs/puppetlabs-stdlib/pull/1298

manifests/fragment.pp
baurmatt: Puppet 5 is still a thing? :D But, yes. It seems Sensitive was already support in Puppet 5 -> https://www.puppet.com/docs/puppet/5.5/lang_data_sensitive.html

ekohl: The reason I mention it is that's what the `if` statement above is about.

b4ldr: the module claims to not support [puppet < 6](https://github.com/puppetlabs/puppetlabs-concat/blob/main/metadata.json#L105) so i wonder if we just drop that if statement

baurmatt: Have nice one! :) I love to remove code I've added ages ago :) brb

ekohl:
I think dropping Puppet 5 compatibility code deserves at least a mention in the changelog. I'd suggest to submit that as its own PR for visiblity and then rebase this once it's merged.

baurmatt:
I don't think it's worth the trouble creating two separate MRs, rebased, ... Officially, Puppet 5 support has been drop with https://github.com/puppetlabs/puppetlabs-concat/pull/685. This just manifests it.

lib/puppet/type/concat_file.rb
```diff
@@ -295,7 +295,11 @@ def nested_merge(hash1, hash2)
 
   def fragment_content(r)
     if r[:content].nil? == false
-      fragment_content = r[:content]
+      fragment_content = if r[:content].is_a?(Puppet::Pops::Types::PSensitiveType::Sensitive)
```
alexjfisher, line 298: I'm a little surprised this is necessary.  I could swear I'm already using `Sensitive` `content` and other than a warning about not being able to automatically redact stuff it's 'working'.  (I still have to manually set `show_diff => false` on the `concat` resource, but no unwrapping anything anywhere.)

alexjfisher: Actually, I'm not.  I'm setting `content` to a `node_encrypt::secret` which is a function that returns a `Deferred`.  The deferred function decrypts the data and _that_ returns a `Sensitive` though. https://github.com/puppetlabs/puppetlabs-node_encrypt/blob/8b1a02f1847e582b58145a0c145f89771e1975bf/lib/puppet/functions/node_decrypt.rb#L12

So I still get the warning `Unable to mark 'content' as sensitive: content is a parameter and not a property, and cannot be automatically redacted.`, but the lacking of unwrapping in the type doesn't break anything for me.

alexjfisher: Given the following manifest
```puppet
concat { '/tmp/testconcat':
  show_diff      => false,
  ensure_newline => true,
}

concat::fragment { 'frag1':
  target  => '/tmp/testconcat',
  content => '# a normal header',
  order   => '01',
}
concat::fragment { 'frag2':
  target  => '/tmp/testconcat',
  content => Sensitive('# my password'),
  order   => '02',
}
concat::fragment { 'frag3':
  target  => '/tmp/testconcat',
  content => '# a normal footer',
  order   => '03',
}
```
`r[:content].class` is apparently a `String` in all 3 cases.

alexjfisher: I'm struggling to replicate the error in spec tests, but the following somewhat convoluted example triggers it.
```puppet
concat { '/tmp/testconcat':
  ensure_newline => true,
}

concat::fragment { 'frag1':
  target  => '/tmp/testconcat',
  content => Deferred('unwrap', ['# some content']),
}
```

```
Error: /Stage[main]/Main/Concat[/tmp/testconcat]/Concat_file[/tmp/testconcat]: Failed to generate additional resources using 'eval_generate': can't modify frozen String
/home/halexfis/concat_test2/concat/lib/puppet/type/concat_file.rb:317:in `fragment_content'
/home/halexfis/concat_test2/concat/lib/puppet/type/concat_file.rb:212:in `block in should_content'
/home/halexfis/concat_test2/concat/lib/puppet/type/concat_file.rb:211:in `each'
/home/halexfis/concat_test2/concat/lib/puppet/type/concat_file.rb:211:in `should_content'
/home/halexfis/concat_test2/concat/lib/puppet/type/concat_file.rb:356:in `eval_generate'
/home/halexfis/.rvm/gems/ruby-2.5.3/gems/puppet-6.29.0/lib/puppet/transaction/additional_resource_generator.rb:56:in `eval_generate'
```

Since with this change to support `sensitive` directly it will probably be much more likely, maybe you want to include a fix for this??

```ruby
fragment_content = fragment_content.dup << newline unless %r{#{newline}\Z}.match?(fragment_content)
```

instead of
```ruby
fragment_content << newline unless %r{#{newline}\Z}.match?(fragment_content)
```
should do it.

LukasAud:
Hey @baurmatt, thanks for taking the time to contribute to our module. Regarding the Puppet 5 code removal topic, while it is true that it has been deprecated for a while, we still prefer (where possible) to have changes properly separated and documented (title and description) in different PRs. This helps us and the rest of the community keep better track of changes being made. Otherwise, pinpointing certain changes can often become a quite challenging.

We would appreciate if you could push the Puppet 5 code removal changes in a different PR. Sorry for the inconvenience.

LukasAud:
#761 has been merged. As soon as this one is rebased and conflicts are resolved, it should be good to merge too.

alexjfisher:
I'm still not convinced the `r[:content].unwrap` line is reachable though.  In other types, where I've had a property/parameter that was a hash and within _that_ there's a sensitive, I'd have to unwrap it, but otherwise I think it's automagic???

lib/puppet/type/concat_file.rb
baurmatt: I've added another commit and attributed to you :)

baurmatt:
Sorry, got sick :( Yeah, seems like this isn't needed. I've removed the `unwrap` part.

LukasAud (approved):
LGTM

natemccurdy:
Thanks for adding Sensitive support! But it appears that the entire `content` has to be wrapped in `Sensitive`. It's not possible to use `epp()` to render the content where certain variables contain `Sensitive` values.

Opened a bug here: https://tickets.puppetlabs.com/browse/MODULES-11429

For example, this doesn't work as expected. A warning is thrown and the sensitive content is not redacted:
```puppet
concat::fragment { 'foo':
  target  => $foo,
  order   => '01',
  content => epp('module/template.epp',
    {
      'password' => Sensitive($password),
      'other'.   => $other,
    }
  ),
}
```

> Warning: /Concat_fragment[foo]: Unable to mark 'content' as sensitive: content is a parameter and not a property, and cannot be automatically redacted.

baurmatt:
@natemccurdy Thanks for bringing this up! I wasn't able to reproduce this with version 7.4.0, but with 8.0.0. So something must have been broken with the latest version.

(Talking about concat version, not Puppet)

baurmatt:
This seems to be broken with https://github.com/puppetlabs/puppetlabs-concat/commit/bc6478ae57ac25be6a53c8c64eca179853c3a598#diff-176bed02b2636bba558658d6d610f63f1c894e45cfa6d7b54e8257cc57ba5d49L98

Ping @LukasAud :)

alexjfisher:
oh, you *definitely* can't just rename that function just because rubocop doesn't like it beginning with `set_`...

alexjfisher:
I'll open a PR.

alexjfisher:
Fixed in https://github.com/puppetlabs/puppetlabs-concat/pull/777

@LukasAud Are there any other modules where a similar change might have slipped in?

alexjfisher:
Maybe remove this https://github.com/puppetlabs/puppetlabs-vcsrepo/blob/main/.rubocop_todo.yml#L45 and explicitly disable the cop on the offending line

ekohl:
> @LukasAud Are there any other modules where a similar change might have slipped in?

https://github.com/puppetlabs/puppetlabs-postgresql/pull/1433 was exactly the same thing.

alexjfisher:
https://github.com/puppetlabs/puppetlabs-chocolatey/blob/bc40015f36960b241651953eaa9aa832a183cbcc/.rubocop_todo.yml#L59 also maybe an accident waiting to happen

alexjfisher:
Good news is I've not been able to find any other modules that have the breaking change in them.

Issue #758, opened 2023-03-11
v7.3.2 change creates error with running `puppet generate types` on Puppetserver


canihavethisone:
## Describe the Bug
Running `puppet generate types` with v7.3.2 of this module creates the error
```
Error: Failed to load custom type 'concat_file' from '/etc/puppetlabs/code/environments/production/modules/concat/lib/puppet/type/concat_file.rb': uninitialized constant Puppet::Type::File
```

## Expected Behavior
Generate types to succeed as it did with previous versions

## Steps to Reproduce
1. install module on Puppetserver
2. run `puppet generate types`

## Environment
 - Puppet version 6 & 7
 - CentOS 7 & 8

## Additional Context
Issue seems to be caused by the removal of one of more of these lines from `lib/puppet/type/concat_file.rb` in the 7.3.2 release
```
require 'puppet/type/file/owner'
require 'puppet/type/file/group'
require 'puppet/type/file/mode'
```

canihavethisone:
@LukasAud bringing this to your attention as the last merge was yours. Thanks!

TheMeier:
I am experiencing the same issue. Had to downgrade concat

canihavethisone:
Thanks for addressing this so quickly!

PR #755, opened 2023-03-08
(MODULES-3522) Removing redundant 'requires'

LukasAud:
Prior to this commit, there was an issue regarding the loading of the puppet/type/file/owner constant under certain circumstances during agent runs.

This commit aims to modernise the file request to comply with best practices.

PR #759, opened 2023-03-13
Revert "(MODULES-3522) Removing redundant 'requires'"

LukasAud:
As established in #758, this update caused an unexpected regression. Reverts puppetlabs/puppetlabs-concat#755

### DIFF
--- a/lib/puppet/type/concat_file.rb
+++ b/lib/puppet/type/concat_file.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require 'puppet/type/file/owner'
+require 'puppet/type/file/group'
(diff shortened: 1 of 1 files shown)


chelnak:
Thanks!

I'd be interested to know exactly why this happened and if it is something that we can test for?

david22swan (approved):
LGTM

canihavethisone:
> Thanks!
> 
> I'd be interested to know exactly why this happened and if it is something that we can test for?

In this case running `puppet generate types` would have detected the issue.  Is it worth adding that to acceptance tests?

ekohl:
I thought about the same thing, though I wonder if it should be an acceptance test or part of the unit test suite.

PR #761, opened 2023-03-15
puppet5: drop remnants of puppet5 code

b4ldr:
puppet 5 support was dropped sometime ago.  This PR simply removes some old code which was around to support puppet5's lack of deferred functions


LukasAud (approved):
LGTM

PR #766, opened 2023-03-29
Add parameter to not create empty files when no fragments are defined

JonasVerhofste:
By default concat would always create an empty file if no fragment was defined. Sometimes we don't want a file unless it has content.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -246,7 +246,7 @@ The following parameters are available in the `concat::fragment` defined type:
 
 ##### <a name="-concat--fragment--content"></a>`content`
 
-Data type: `Optional[Any]`
+Data type: `Optional[Variant[String, Deferred]]`
 
 Supplies the content of the fragment. Note: You must supply either a content parameter or a source parameter.
 Allows a String or a Deferred function which returns a String.
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -56,6 +56,7 @@ The following parameters are available in the `concat` defined type:
 * [`show_diff`](#-concat--show_diff)
 * [`validate_cmd`](#-concat--validate_cmd)
 * [`warn`](#-concat--warn)
+* [`create_empty_file`](#-concat--create_empty_file)
 
 ##### <a name="-concat--backup"></a>`backup`
 
@@ -222,6 +223,14 @@ it explicitly if you need it.
 
 Default value: `false`
 
+##### <a name="-concat--create_empty_file"></a>`create_empty_file`
+
+Data type: `Boolean`
+
+Specifies whether to create an empty file if no fragments are defined. Defaults to true.
+
+Default value: `true`
+
 ### <a name="concat--fragment"></a>`concat::fragment`
 
 Manages a fragment of text to be compiled into a file.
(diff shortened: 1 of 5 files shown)


JonasVerhofste:
Ran a rubocop:auto_correct, should be good now.

Also, this is a non-breaking change. Default behaviour is how it was before the param existed.

LukasAud (approved):
LGTM

JonasVerhofste:
@elofu17 This sounds very much alike #772  (see the mention above your comment).

The module you mention has the same "outdated" version requirement (i.e. lower than 8).

Could you test with puppetlabs/concat v7.4.0? That version should have the new param as well but should work with given the metadata of saz/ssh v10.1.0. If that works, then metadata is here the issue as well. :)

elofu17:
I will upgrade concat from 7.3.0 to 7.4.0 now and see.

elofu17:
Identical error with 7.4.0.
I'm reverting back to 7.3.0 again. :-)

siebrand:
Same issue here with puppet-nginx 4.4.0 and puppetlabs-concat 8.0.0, as well as with puppetlabs-concat 7.4.0.

`Server Error: no parameter named 'create_empty_file' (file: /etc/puppetlabs/code/environments/production/modules/concat/manifests/init.pp, line: 126) on
Concat_file[/etc/nginx/sites-available/00-http-php-80.conf] (file: /etc/puppetlabs/code/environments/production/modules/concat/manifests/init.pp, line: 126)`

Highest version that works is concat 7.3.3.

JonasVerhofste:
@siebrand @elofu17 Just to confirm, you did run type generation (`puppet generate types --environment ${env}`) after bumping the module, right? Because this looks a lot like the manifest being updated, but puppet not correctly knowing about the new param in the type.

elofu17:
Nope. :-) I don't even know what you are saying to be honest. :)

I've always just bumped the module version number in my Puppetfile.
On the next r10k run on the puppet-server, the environment(s) get the new 3rd party modules downloaded and populated in the .

Then a client node performs a puppet run and the error is triggered.

I.e. I don't restart the puppet server, I don't clear its caches, I don't generate types or anything. Is this wrong?

elofu17:
I just tested upgrading to v8.0.0 again, and then restarted the puppet server.
Now it works :-)

It sounds like I was using some cached version instead of the new.
Interesting that this is the first time I notice any problems. :) Sure, some of the times when I bump the 3rd party modules I also upgrade the puppet server itself, and hence restart it == those times I wouldn't have this kind of problem. But I know that I have also bumped modules without a restart lots of times during the years, and this is the first time I have noticed a problem.

I'll remember to restart the server (or clear its cache) upon bumping modules in the future.

Sorry to have wasted your time, and thanks for your comments.

JonasVerhofste:
@elofu17 no worries, most important part is that it's fixed! A colleague pointed out that r10k probably takes care of the type generation, so I think you're good. (We're not using r10k, which is why we need to run the type generation manually)

@siebrand could you test bumping and then restarting puppetserver as well?

siebrand:
@JonasVerhofste I can confirm that the issue with 7.4.0/8.0.0 does not surface after a restart of puppetserver. Thank you for the help.

PR #774, opened 2023-04-19
(CONT-891) Address nightly rubocop failures

LukasAud:
Prior to this commit, a rubocop failure appeared in the nightly CI. This rule was not picked up during the Puppet 8 update for unknown reasons. This commit aims to troubleshoot that rubocop rule so that CI can resume normal operations.


jordanbreen28 (approved):
LGTM

PR #776, opened 2023-05-04
(MAINT) Addressing wrong Rubocop TargetRubyVersion

LukasAud:
Prior to this commit, during the Puppet 8 support work, TargetRubyVersion was set to 2.7 and a cop rule was addressed according to it. However, the correct ruby version is 2.6.

This commit aims to address the previously wrongly set version along with an unnecesary change that breaks some compatibility with 2.6 systems.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,7 +7,7 @@ require:
 AllCops:
   DisplayCopNames: true
   NewCops: enable
-  TargetRubyVersion: '2.7'
+  TargetRubyVersion: '2.6'
   ExtraDetails: true
   DisplayStyleGuide: true
   Include:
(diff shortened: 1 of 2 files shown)


ekohl (approved):
I think this will also reintroduce compatibility for Puppet 6 users, which is nice.

jordanbreen28 (approved):

alexjfisher:
Worth noting that 2.6 was the correct target version as that's the version compatible with Puppetserver 7's jruby.

alexjfisher:
@bastelfreak Can we fix https://github.com/voxpupuli/voxpupuli-rubocop/issues/16 now?

PR #777, opened 2023-05-05
Revert "Correct Naming/AccessorMethodName"

alexjfisher:
This reverts commit bc6478ae57ac25be6a53c8c64eca179853c3a598.

The API is `sensitive_parameters`, we can't just change the name of this method in this type because rubocop complains.


ekohl (approved):

natemccurdy:
Thanks for the fast turn around @alexjfisher 

Though I wonder if there's an rspec test that could be added to catch a regression in expected behavior for the future.

I could try to add one later today when I find the time.

LukasAud (approved):
Ah yes, this is another revert I was planning to take a look at. LGTM

Issue #775, opened 2023-05-03
Concat 8.0.0 / Puppet 8.0 - Deferred function calls incompatibility


fetzerms:
## Describe the Bug
Concat v8 (and also prior) does not play well with Deferred function calls in Puppet 8. They do not get properly evaluated when using them as a content parameter. I first encountered this issue at https://github.com/voxpupuli/puppet-gitlab_ci_runner/issues/167 - but tracked it down to concat / puppet v8.

## Expected Behavior
Concat should be able to properly use Deferred function calls.

## Steps to Reproduce
With docker. 

Filename: Dockerfile-8
```
FROM debian:bullseye
RUN apt-get update && apt-get install -y wget
RUN wget https://apt.puppetlabs.com/puppet8-release-bullseye.deb
RUN dpkg -i puppet8-release-bullseye.deb
RUN apt-get update && apt-get install -y puppet-agent
RUN /opt/puppetlabs/bin/puppet module install puppetlabs-concat --version 8.0.1
RUN /opt/puppetlabs/bin/puppet apply -e "concat { '/tmp/file': ensure => present, } concat::fragment { 'tmpfile': target  => '/tmp/file', content => Deferred('md5', ['Hello']), order   => '01'}"
```

Commands:
`docker build -f Dockerfile-8 -t concat8_puppet8 . && docker run -it concat8_puppet8 cat /tmp/file`

Output:

```
[+] Building 13.1s (11/11) FINISHED                                                                                                                                                                                                                                                                                                                                        
 => [internal] load build definition from Dockerfile-8                                                                                                                                                                                                                                                                                                                0.0s
 => => transferring dockerfile: 543B                                                                                                                                                                                                                                                                                                                                  0.0s
 => [internal] load .dockerignore                                                                                                                                                                                                                                                                                                                                     0.0s
 => => transferring context: 2B                                                                                                                                                                                                                                                                                                                                       0.0s
 => [internal] load metadata for docker.io/library/debian:bullseye                                                                                                                                                                                                                                                                                                    0.5s
 => [1/7] FROM docker.io/library/debian:bullseye@sha256:63d62ae233b588d6b426b7b072d79d1306bfd02a72bff1fc045b8511cc89ee09                                                                                                                                                                                                                                              0.0s
 => CACHED [2/7] RUN apt-get update && apt-get install -y wget                                                                                                                                                                                                                                                                                                        0.0s
[...]

PR #789, opened 2023-07-19
#775 - Explicitly resolve deferred values

fetzerms:
From Puppet Agent v8 onwards, a deferred value needs to be explicitly resolved.

## Summary
This adds explicitly resolving of deferred values. See #775 
If that fix is suitable (works for me), I'll be happy to add Spec and acceptance tests for deferred values.

## Additional Context
I don't know of the root cause. Seems Puppet Agent v8 is doing things differently than <=v7

## Related Issues (if any)
#775 

## Checklist
- [ ] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [x] Manually verified. (For example `puppet apply`)


smortex (approved):
LGTM!

smortex:
Failed unit tests seems transient. Closing and re-opening to trigger a new run.

fetzerms:
@smortex - As the build is green now, do we think we can merge this?

smortex:
That seems good to me. There is no "policy" AFAIC for external contributors like me and other folks that where added as reviewers, but I would prefer to see a review/approval by someone from @puppetlabs since they own the project.

joshcooper (approved):
I filed a puppet bug about this, the intention was not to leak deferred values into custom type/providers. But this seems like an ok workaround.

bastelfreak (approved):

PR #844, opened 2026-09-17
Add deduplicate_fragments parameter to concat

### LINKED ISSUE #839: Disallow duplicates
## Use Case
I'm using fragments in a define and I would like to disallow duplicates going into a file. Example if I have a two different calls doing a fragment with content 10\n I get two even though I only want one.

## Describe the Solution You Would Like
On the main concat I would like to see something like allow_duplicate => false and the default would be true. This wouldn't cause an error in puppet and the run to fail but it would just skip any duplicates. In my example there would only be one 10\n

## Describe Alternatives You've Considered
A way to query the concat in an if statement or unless, if ! concat{"test"}["10\n"] { run the fragment } I just put that together on the fly but I would like to use an if statement or unless to skip the fragment

## Additional Context
While I'm making this work with the processing of the file after puppet I would like to see the ability to keep duplicates from being added.

jst-cyr:
## Summary

- Adds a new opt-in `deduplicate_fragments` boolean parameter (default `false`) to the `concat` defined type and the `concat_file` native type.
- When enabled and `format => plain` (the default), fragments whose content exactly duplicates a fragment already emitted earlier in the sorted order are skipped — the fragment with the lowest `order` (or, for ties, the alphabetically/numerically first name) wins.
- Has no effect for the `yaml`/`json*` formats, since those already deduplicate merged values by construction (duplicate hash values under the same key don't error, and array merges already run `.uniq` unconditionally).
- Fragments with `Sensitive`-wrapped content are exempt from the comparison, so secret values are never compared for equality or silently collapsed.
- Default is unchanged behavior — fully backward compatible.

Addresses #839.

## Test plan

- [x] `bundle exec rake syntax lint metadata_lint rubocop` — clean
- [x] `bundle exec rake parallel_spec` — 268 examples, 0 failures, including new `#should_content` unit tests covering the default (off), dedup-on, order tie-break, distinct-content, and sensitive-content-not-collapsed cases
- [x] `bundle exec rake strings:generate:reference` — REFERENCE.md regenerated
- [x] `spec/acceptance/deduplicate_fragments_spec.rb` — new Litmus acceptance spec added; needs a provisioned target to run (not exercised locally in this environment)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jst-cyr:
Failures on AlmaLinux 8 match failures in main.
Initially failed on Windows environments, but handling of new line characters has been corrected and additional test cases to ensure Windows line endings are handled properly.

